PR #7, opened 2015-04-17
Fix doc comment

rsaarelm:
Looks like some collateral damage from an int -> i32 search and replace.


mattyhall:
Thanks very much. I should be more careful with my vim use! :smile:

PR #9, opened 2015-04-30
modified integers to take floats instead

sinistersnare:
Hey, my code works now :)

If theres any other pieces of this you would like to change, let me know, but i think this is ok.

I did find that you do not allow `Ellipse`s to have 0 width or height. What was the decision on that? Tiled seems fine with it.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -372,12 +372,12 @@ impl ObjectGroup {
     }
 }
 
-#[derive(Debug, PartialEq, Eq)]
+#[derive(Debug, PartialEq)]
 pub enum Object {
-     Rect { x: i32,  y: i32,  width: u32,  height: u32,  visible: bool},
-     Ellipse { x: i32,  y: i32,  width: u32,  height: u32,  visible: bool},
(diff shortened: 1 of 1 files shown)


mattyhall:
No, that's great thanks. I'm glad it's working for you!

>I did find that you do not allow Ellipses to have 0 width or height. What was the decision on that? Tiled seems fine with it.

To be honest, I can't remember. I guess at the time I thought it doesn't make any sense to have objects that have zero area. I suppose there are a few usecases though - you might just want to flag a particular position or something.

It's something that probably needs changing. Do you use this feature in particular? Do you need it for rects as well?

sinistersnare:
Yeah, I decided to use an ellipse for a spawn point, and by default Tiled makes it 0 sized, so it seemed like it should be a possibility if its the default.

Right now I am not using rectangles, if Tiled allows it, it should probably be allowed. lemme see if its a thing there.

sinistersnare:
Yeah, it seems that 0 sized objects are generally allowed, Ill see if theres a change to make. I see you made the Ellipse change too , thanks, I was on the run yesterday :)

PR #10, opened 2015-05-02
Piston Renderer example

sinistersnare:
Hey, I thought it would be nice to include an example of drawing a tiled map using Piston, so I took a few minutes to write it out.

I dont have any formal experience in game dev, ive just been noodling around for a few years, so if I am doing evil game-dev things please let me know :p

Drawing objects is not supported currently, but it is being worked on.

It can be merged as is, if you want it, but if you do not, have suggestions, or want to wait until more is done, please let me know :)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,3 +2,4 @@ tags
 target
 Cargo.lock
 *.swp
+*.dll
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,6 +2,11 @@ env:
     global:
         - secure: zpxgR0Nn7YhCRXPjfR/gcdxAObeJLBoRMW0qeKiyuc4V6Y1guZ9EfnvJ5uL4pRXnP0LZrXm/WwYicXm7J0vTwQwq/4Bf0NXykysZfVrHFTY5Wook/QXJd8OajCwuJa8d/wr6Dx/wGe/80gi8hZibraMS5QkWKgtzz9HCFLM99uw=
 language: rust
+before_install:
+  - sudo add-apt-repository ppa:team-xbmc/ppa -y
+  - sudo apt-get update -qq
+install:
+  - sudo apt-get install libsdl2-dev
 script:
     - cargo build
     - cargo test
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -17,7 +17,26 @@ path = "src/lib.rs"
 name = "example"
 path = "examples/main.rs"
 
+[[example]]
+name = "piston"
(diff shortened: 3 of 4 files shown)


mattyhall:
This is great, you're a star! I looked into Piston when I first wrote this but I couldn't find any methods to draw a bit of an image (like ```drawn_image.src_rect```).

There's just a few things:

  * Are all the piston crates on crates.io? As Rust 1.0 is soon I changed ``Cargo.toml`` to use crates.io packages only - not git - and removed an unstable feature so that users did not have to build on nightly.
  * Could you change the piston stuff to a [dev dependency](https://github.com/rust-lang/cargo/blob/master/src/doc/manifest.md#the-dev-dependencies-sections). This just means piston won't have to be built if someone wants to used the library with a different game engine/framework/whatever.

Thanks again!

sinistersnare:
The crates are on crates.io, but they are not updated, as some deps are git only. So that wont work at the moment.

They are currently using dev dependencies, so they are only built when running `cargo test` or `cargo run --example piston`

So if they want to see the piston example, they will possibly need to be on nightly (I dont remember if all the deps are on beta, but I think they are), but because they are dev dependencies, they dont need nightly to compile the library itself!

I hope thats an ok compromise. Ill bug people on PistonDevelopers to update their deps and see if we can eventually move to a crates.io solution.

mattyhall:
Really sorry, I didn't notice they were dev dependencies. I need my eyes testing! :smile: 

When 1.0 comes around I'll be sort out the ``Cargo.toml`` and put some versions on the dependencies.

The Travis build failed, but everything builds on my machine. Perhaps you could install SDL with apt instead? [gfx_graphics](https://github.com/PistonDevelopers/gfx_graphics/blob/master/.travis.yml) by the Piston devs does that.

sinistersnare:
I followed a different .travis.yml, but it didnt work! So I just pushed what the gfx_grahpics guys do (but i didnt add the xbmc ppa, not sure how that would help?) 

Lets see!

sinistersnare:
It works! Anything else you need?

mattyhall:
That's great thanks. Would you mind squashing the commits into one, please? Just to keep the history clean.

Thanks again for all your work. It's much appreciated.

sinistersnare:
Ok, I actually did it correctly :dancer: You can let travis go one more time i guess, thanks for the input! Glad this works, and I hope it helps people looking for examples in the future.

mattyhall:
Cool! All sorted! Cheers

PR #11, opened 2015-10-02
Fix dependencies to make the main library compile

rsaarelm:
The Piston example being built with 'cargo test' is also broken, probably needs a more detailed look. This fix should be enough to keep the library usable as a dependency.


mattyhall:
Thanks very much for the fix.

PR #12, opened 2015-10-02
Fix markdown formatting in CONTRIBUTORS file

rsaarelm:
The old version globs all names into a single paragraph when rendered to markdown.


alexdevteam:
LGTM. Not sure if we should keep the CONTRIBUTORS file though (either way we'd need to update it)

PR #13, opened 2015-11-01
Update for rust v1.4

axelmagn:
I did a quick once over to git this package building and passing tests again.  This included removing the piston example, since the piston library has changed substantially since then and I didn't entirely know how to fix it.  I didn't see it as essential to the package, and if anyone wants to reimplement it they can always check the git logs.

 * Set all package dependencies to fixed version numbers
 * Removed outdated piston dev dependencies
 * Removed piston example
 * Updated travis.yml
 * Ticked version to 0.1.4


mattyhall:
Thanks very much. Sorry for the delay in merging.

PR #15, opened 2016-01-01
Edited orientation keywords to be compatible with Tiled 0.14.2

hegza:
I'm not sure whether it was a bug in an older Tiled version that the 'Staggered' keyword was capitalized, it's definitely inconsistent. Tiled 0.14.2 is a daily build in which that issue's been fixed, apparently.


mattyhall:
Cheers

PR #18, opened 2016-07-10
Add Error trait implementation for TiledError.

### LINKED ISSUE #17: TiledError doesn't implement Error
This prevents it from having the semantics one would expect of an error type.

SilverWingedSeraph:
This makes TiledError behavior more standard, more expressive, and much
more useful for both library and application developers. Close #17


mattyhall:
Ta very much

PR #19, opened 2016-07-23
Added tile id to image, since they don't have to start at 0

RubenSlabbert:
Images in `tileset` generally start at an id of 0, but if you remove all the images from a tileset and then add new ones, without recreating the tileset, they'll start where they left off, but the firstgid can still be one. For example, the situation that I discovered:
```xml
<tileset firstgid="1" name="Floors" tilewidth="32" tileheight="32" tilecount="18" columns="0">
  <tile id="54">
     <image width="32" height="32" source="../sprites/floors/experimentFloor1.png"/>
  </tile>
```
This commit adds an id to the Image struct which allows the user facing code to avoid this situation.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -297,7 +297,7 @@ impl Tileset {
 
         let mut images = Vec::new();
         parse_tag!(parser, "tileset",
-                   "image" => |attrs| {
+                   "tile" => |attrs| {
                         images.push(try!(Image::new(parser, attrs)));
                         Ok(())
                    });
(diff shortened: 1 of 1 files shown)


RubenSlabbert:
Looking at the tests it seems like there's a difference between using a tileset to contain a single spritesheet and to contain multiple images.

RubenSlabbert:
Alright, the tests pass now. I didn't implement all the tile attributes, since Terrain isn't support by the library yet, and probability is a float, which doesn't support `Eq` and would break the rest of the tree.

Also, the Readme seems to be outdated since judging by the code the library does support files that aren't base64 encoded and gzipped.

mattyhall:
Thanks very much  :smile: You're right about the README, I'll update it.

Cheers!

PR #20, opened 2016-09-13
Add missing object attributes

standart0815:
The tmx format supports for objects various attributes.
id, gid, name and type were ignored by the parser.
as an result it was not possible not possible to have objects which
are using tiles because of the missing attribute gid.

The enhanced parser is now parsing the additional attributes:
*id  - object id,
*gid - tile no
*name- object name (property)
*type- object type (property)


mattyhall:
Cheers :)

PR #21, opened 2017-02-04
Properly parse object types

mystal:
The attribute is called "type" but we were previously looking for
"obj_type".


mattyhall:
Cheers :+1:

PR #22, opened 2017-03-19
Add properties to Objects and parse properties as enums

mystal:
I needed to read properties from objects and figured I'd clean up some things while I was adding support for that :)

Mostly made properties type safe (I don't support all variants yet, but at least all the ones being used in tests). I also split out object shapes to reduce redundancy in the Object struct.


mattyhall:
Sorry for not getting back to you sooner. This is great, thanks!

mystal:
2 day turnaround is plenty quick for open source :)

Cheers!

PR #23, opened 2017-05-13
Add keywords metadata to Cargo.toml

afonso360:
Hi, it took me a while to find this crate in crates.io, I had used tiled (the editor) before and so searched for both tmx and maps but found nothing.

I think this would improve visibility for this crate


mattyhall:
Thanks :)

PR #24, opened 2017-06-09
Load Tilesets From External Files

guodman:
When the tileset is in an external file, parse that file when given the path of the map file.  To load an extenal tileset, the new parse_file function must be used.  If the map has an external tileset and the original parse endpoint is used, the result will be an error.


Phrohdoh:
Would you be opposed to publicly exporting TSX parsing, too?

I could use that in a project of mine.

guodman:
That should be easy to do.  I'll work on that this weekend and update this pull request.

src/lib.rs
```diff
@@ -351,7 +358,62 @@ impl Tileset {
...
+
+    fn parse_external_tileset<R: Read>(g: u32, parser: &mut EventReader<R>, attrs: &Vec<OwnedAttribute>) -> Result<Tileset, TiledError> {
+        let ((s, m), (n, w, h)) = get_attrs!(
+            attrs,
+            optionals: [("spacing", spacing, |v:String| v.parse().ok()),
+                        ("margin", margin, |v:String| v.parse().ok())],
+            required: [("name", name, |v| Some(v)),
+                       ("tilewidth", width, |v:String| v.parse().ok()),
+                       ("tileheight", height, |v:String| v.parse().ok())],
+            TiledError::MalformedAttributes("tileset must have a firstgid, name tile width and height with correct types".to_string()));
```
Phrohdoh: `firstgid, name, tile, width, and height` is easier to comprehend.

guodman:
I have updated this PR with a public method for parsing just a tsx file, as well as the other feedback you gave.  Thanks for the suggestions.

mattyhall:
Sorry for the delay. This is really cool, cheers!

PR #27, opened 2017-06-14
Load in properties from the tiles

guodman:
I had to remove Eq from the Tile and Tileset due to the fact that f32 does not implement it.


mattyhall:
Sorry for the delay, and thanks! It's a bit annoying rust does this, but oh well!

PR #32, opened 2017-08-10
Tilesets can have objects inside of them since Tiled 1.0,

Connicpu:
This just adds a field to the Tile so that data may be parsed and accessed

Also, the XML may omit the name for these objectgroups, so I made ObjectGroup::name default to an empty string


mattyhall:
Nice, thanks very much!

PR #36, opened 2017-10-21
Parse rotation on entities

Nnubes256:
The [TMX standard](http://doc.mapeditor.org/en/latest/reference/tmx-map-format/) defines the rotation property on objects as:
```
- rotation: The rotation of the object in degrees clockwise (defaults to 0).
```
The current code ignores this property.

A small note: object rotation pivots around the object's **top left corner**, not around the center(see https://github.com/bjorn/tiled/issues/539#issuecomment-40653258).

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -578,6 +578,7 @@ pub struct Object {
     pub obj_type: String,
     pub x: f32,
     pub y: f32,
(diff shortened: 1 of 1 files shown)


mattyhall:
Thanks

PR #39, opened 2017-12-24
Parse imagelayer

TatriX:
Also add test for parsed image layers.
The test can be also used as an example off how one can get data from
the parsed map.


mattyhall:
Thanks very much :)

PR #44, opened 2018-06-03
Add tile-flipping support

Jengamon:
Not sure if the method is optimal, but here is some byte-munchin', tile-flippin' goodness.


mattyhall:
Thanks

Issue #34, opened 2017-09-20
Layering information between different types of layers can't be used


LaylBongers:
Tiled lets you layer different types of layers in whatever way you want:
![](https://i.imgur.com/HVLRfKo.png)

This is useful to for example define in what layer of a map the enemies are, so you can have tiles that are rendered above enemies. You can also have different layers of enemies (flying enemies) or effects (moving cloud, flying car entities).

This library however doesn't keep track of that information, the different layer types are sorted into their own vectors:
![](https://i.imgur.com/sACOLuf.png)

It would be better to have one single list of layers instead, with every layer entry being an enum that defines what type of layer it is, or a separate list that defines the full layering.

bjorn:
> It would be better to have one single list of layers instead, with every layer entry being an enum that defines what type of layer it is, or a separate list that defines the full layering.

Yes, I think that'd be the best approach, but in the meantime at least #51 provided a solution for this issue.

PR #51, opened 2018-12-17
Keep track of layer indexing and parse more tile data

### LINKED ISSUE #45: Tile missing type property
Title says it all. The Tile struct provided does not include the type field. See the [TMX Map Format](http://docs.mapeditor.org/en/stable/reference/tmx-map-format/#tile) for details on what it is.

Frizi:
This should fix #45 and #34.

Also @mattyhall, could you go through existing issues and check on them? Most of those are already fixed, so it's just a metter of closing them.


mattyhall:
Thanks

PR #52, opened 2018-12-18
Rustfmt pass and slight macro stylistic changes

Frizi:
Everything I did was to run a rustfmt over the code and fix some misformatted macros by hand. To do that i had to slightly modify the macro syntax to add a wrapping block for tag matching and to allow trailing commas where appropriate.

I also see a potential to remove those macros entirely and just do it with normal functions (or at least remove unnecessary parts of those macros), but that's a change for another commit.

### DIFF
--- a/examples/main.rs
+++ b/examples/main.rs
@@ -1,7 +1,7 @@
 extern crate tiled;
 
-use std::path::Path;
 use std::fs::File;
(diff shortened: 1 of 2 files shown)


mattyhall:
Thanks, it looks much better now

PR #53, opened 2019-04-08
Support for groups

agmcleod:
Hi there,

I started on a game project, wanting to use groups to sorta simulate a scene graph type representation. Given rs-tiled didn't have that, I thought I'd implement and submit a PR. Here's a summary of the changes:

* Breaking change - renamed `Layer` struct to `TileLayer`. Reason for this, is I needed a trait type that each layer type implemented, and I thought Layer was the more generic name. If you wish to avoid the breaking change, I can name it to something else. Definitely open to suggestions here
* Added the downcast crate, so the `Group` struct can store a `Vec<Layer>` and downcast into the specific types
* Some tests and code just got reformatted, as my editor uses fmt on save.
* Only thing I don't like is the bit of duplicate code in `parse_tag!` in the `Group::new` function. Which is fairly re-purposed from `Map::new`. If you have suggestions on how to refactor this, I am also open to suggestions. Though it's a little tricky, given that in `Group::new` the layers are pushed as a Box type to a single Vec instead of across many.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -21,5 +21,6 @@ path = "examples/main.rs"
 
 [dependencies]
 base64  = "0.5.2"
+downcast-rs = "1.0.3"
 xml-rs  = "0.3.0"
 libflate = "0.1.18"
--- a/assets/tiled_group_csv.tmx
+++ b/assets/tiled_group_csv.tmx
@@ -1,0 +1,44 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<map version="1.2" tiledversion="1.2.2" orientation="orthogonal" renderorder="right-down" width="10" height="10" tilewidth="32" tileheight="32" infinite="0" nextlayerid="8" nextobjectid="3">
+ <tileset firstgid="1" name="tilesheet" tilewidth="32" tileheight="32" tilecount="84" columns="14">
+  <image source="tilesheet.png" width="448" height="192"/>
+ </tileset>
+ <layer id="1" name="Tile Layer 1" width="10" height="10">
+  <data encoding="csv">
+43,43,43,43,43,43,43,43,43,43,
+17,6,7,7,7,7,7,7,8,17,
+17,20,21,21,21,21,21,21,22,17,
+17,20,21,21,21,21,21,21,22,17,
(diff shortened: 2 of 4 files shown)


mattyhall:
Hey, sorry this took so long. As I hope you can appreciate life often gets in the way of open source. Particularly when I'm so forgetful! To answer your points:

* I think I'd prefer not renaming the Layer struct. It seems simpler to me to keep the names the same as the tags. Of course naming things is the hardest problem so I don't know what the name of the trait should be!
*  I've never come across the downcast crate before. I think if I had been writing it I would have used an enum. Not sure which is better - I might have to have a think about that.

I think the Layer trait (as it's called now) could have at least an id function..? In general empty traits like that _feel_ wrong to me. There's no real reason why I think they're bad (I know they're used elsewhere in rust) but if we can have some functionality there then that's all to the good.

Once again, apologies for being so slow (and, with my last merge, allowing this to bit rot).

paaaav:
Now that the latest version of Phaser 3 supports groups too I'd love to see support of groups in rs-tiled as well.

agmcleod:
ah i should really come back to this and address the changes. I recall seeing the comments, but had a real busy work summer :)

agmcleod:
@mattyhall Very much overdue, but i made some updates and fixed conflicst. I added a function to the trait to return the layer index. I also renamed it to AnyLayer, but I'm still not super happy with this name.

paaaav:
@mattyhall Hey there :) Is there any way to get it in? What needs to be done?

alexdevteam:
Sorry for the delay. Since this PR was opened, this crate is has moved to the official [mapeditor](https://github.com/mapeditor) organization. As such, we are looking for new members to contribute to this crate and review new PRs.

As for the PR itself, I do agree with @mattyhall in terms of the dynamic casts. Since layer types are known at compiletime, dynamic references are not needed here, and enums can work better in this case (Leads to better performance and less runtime errors overall). Apart from that, the crate got reorganized in #104 and as such, some changes are needed. If you can fix these issues I'll be happy to review the PR in detail. Thank you for your time! :)

agmcleod:
@bjorn sounds good! Sorry i wasn't able to update as well :)

PR #56, opened 2019-06-03
Remove broken flipped tile handling

alec-deason:
The current code applies the bit twiddling necessary to decode tile flip status out of tile ids to the ids of tiles inside tilesets but that's not actually where the flips are encoded. As a result the tile ids reported by rs-tiled are incorrect. This change moves that code onto the global ids of tiles inside the layers where the filp state is actually encoded.

This removes the flip_v and flip_h fields from Tile because those were not being set correctly anyway and Tile is not the right location for them. They should be attached to the Layer somehow but this change doesn't propose a way to do that.


alec-deason:
Closing since #61 provides a complete solution rather than just dropping the feature.

PR #58, opened 2019-08-15
Flatten tiles to vec u32

satrix321:
Hi,
This change flattens the tiles from a layer into a structure containing Vec<u32> (instead of Vec<Vec<u32>>) for performance concerns. There's also a method provided for easy conversion to Vec<Vec<u32>> if there would be need for it.

As of now this is a breaking change, but with some minor adjustments users would still be able to iterate over this collection by rows (by using tiles_per_row and number_of_rows). I'm open to any suggestions how we could smooth this out a bit more.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -627,25 +627,15 @@ impl Image {
 pub struct TilesContainer {
     pub data: Vec<u32>,
     pub tiles_per_row: u32,
-}
-
-impl std::ops::Index<usize> for TilesContainer {
-    type Output = [u32];
-
-    fn index(&self, index: usize) -> &Self::Output {
-        let tiles_per_row = self.tiles_per_row as usize;
-        let start_index = index * tiles_per_row;
-        let end_index = start_index + tiles_per_row;
-        &self.data[start_index..end_index]
-    }
+    pub number_of_rows: u32,
 }
 
 impl From<&TilesContainer> for Vec<Vec<u32>> {
     fn from(item: &TilesContainer) -> Self {
         let mut data: Vec<Vec<u32>> = Vec::new();
         let tiles_per_row = item.tiles_per_row as usize;
-        let number_of_vectors = item.data.len() / tiles_per_row;
-        for x in 0..number_of_vectors {
+        let number_of_rows = item.number_of_rows as usize;
+        for x in 0..number_of_rows {
             let start_index = x * tiles_per_row;
             let end_index = start_index + tiles_per_row;
             let row: Vec<u32> = (&item.data[start_index..end_index])
@@ -691,6 +681,7 @@ impl Layer {
         let mut tiles = TilesContainer {
             data: Vec::new(),
             tiles_per_row: 0,
+            number_of_rows: 0,
         };
         let mut properties = HashMap::new();
         parse_tag!(parser, "layer", {
@@ -1105,9 +1096,11 @@ fn decode_csv<R: Read>(
                     .split(",")
                     .map(|v| v.parse().unwrap())
                     .collect();
+                let number_of_rows = data.len() as u32 / width;
                 let tiles = TilesContainer {
                     data,
                     tiles_per_row: width,
+                    number_of_rows,
                 };
                 return Ok(tiles);
             }
@@ -1116,6 +1109,7 @@ fn decode_csv<R: Read>(
                     return Ok(TilesContainer {
                         data: Vec::new(),
                         tiles_per_row: 0,
+                        number_of_rows: 0,
                     });
                 }
             }
@@ -1136,9 +1130,11 @@ fn convert_to_u32(all: &Vec<u8>, width: u32) -> TilesContainer {
             data.push(n);
         }
     }
+    let number_of_rows = data.len() as u32 / width;
     TilesContainer {
         data,
         tiles_per_row: width,
+        number_of_rows,
     }
 }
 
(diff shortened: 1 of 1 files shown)


mattyhall:
Sorry for the delay, I'll try and look at this at the weekend!

mattyhall:
Sorry for the delay again. Life got in the way I'm afraid :( 

It looks good. I'm not too bothered about the breaking version as we're still <1.0 (and I doubt anyone is seriously using the library) but maybe we could add a conversion from TilesContainer to Vec<Vec<_>>

satrix321:
Hey, I've tried looking into this conversion, and it seems to me that we could add a generic one for Vec<Vec<T>>, but it will require a wrapper struct to work properly. Something along those lines:

`struct TilesContainerWrapper<T>(Vec<Vec<T>>);
impl<T> From<&TilesContainer> for TilesContainerWrapper<T> where T: std::convert::TryFrom<u32> {
    fn from(item: &TilesContainer) -> TilesContainerWrapper<T> {
        let mut data: Vec<Vec<T>> = Vec::new();
        let tiles_per_row = item.tiles_per_row as usize;
        let number_of_rows = item.number_of_rows as usize;
        for x in 0..number_of_rows {
            let start_index = x * tiles_per_row;
            let end_index = start_index + tiles_per_row;
            let row: Vec<T> = (&item.data[start_index..end_index])
                .iter()
                .cloned()
                .map(|e| T::try_from(e).ok().unwrap())
                .collect();
            data.push(row);
        }
        TilesContainerWrapper(data)
    }
}`

I'm not exactly sure if this is something desirable in this case. We could just go with implementing the conversion manually for all built in numeric types, but it would produce a lot of boilerplate code. I'm open to any suggestions on how we could improve this solution.

src/lib.rs
```diff
@@ -623,14 +623,39 @@ impl Image {
     }
 }
 
+#[derive(Debug, PartialEq, Clone)]
+pub struct TilesContainer {
+    pub data: Vec<u32>,
+    pub tiles_per_row: u32,
+    pub number_of_rows: u32,
```
bjorn, line 630: Maybe just `width` and `height` would do as property names?

src/lib.rs
```diff
@@ -623,14 +623,39 @@ impl Image {
...
+}
+
+impl From<&TilesContainer> for Vec<Vec<u32>> {
```
bjorn, line 633: I don't think we need this converter. What use would people have for converting to this sub-optimal data structure? Maybe instead it could be useful to provide a getter to retrieve a tile given `x` and `y` coordinates?

bjorn (commented):
This seems like a good change to make. Would you still be able to rebase this change to resolve the conflicts? I've also left some comments.

src/lib.rs
```diff
@@ -653,7 +678,11 @@ impl Layer {
             ],
             TiledError::MalformedAttributes("layer must have a name".to_string())
         );
-        let mut tiles = Vec::new();
+        let mut tiles = TilesContainer {
```
alexdevteam, line 681: Could this be initialized with correct data from the start? I feel like this could lead to invalid states where the tiles member never got initialized.

src/lib.rs
```diff
@@ -1054,49 +1083,59 @@ fn decode_gzip(data: Vec<u8>) -> Result<Vec<u8>, TiledError> {
...
-                    rows.push(
-                        row.split(',')
-                            .filter(|v| v.trim() != "")
-                            .map(|v| v.replace('\r', "").parse().unwrap())
-                            .collect(),
-                    );
-                }
-                return Ok(rows);
+                let data: Vec<u32> = s
+                    .replace('\r', "")
```
alexdevteam, line 1094: I think `trim` after `split` would work here.

src/lib.rs
```diff
@@ -1054,49 +1083,59 @@ fn decode_gzip(data: Vec<u8>) -> Result<Vec<u8>, TiledError> {
...
+                    data,
+                    tiles_per_row: width,
+                    number_of_rows,
+                };
+                return Ok(tiles);
             }
             XmlEvent::EndElement { name, .. } => {
                 if name.local_name == "data" {
-                    return Ok(Vec::new());
+                    return Ok(TilesContainer {
```
alexdevteam, line 1109: If we're going to return a null container I think it would be better to return `None` (or some kind of `Error`) rather than an empty tiles container.

alexdevteam (changes requested):

src/lib.rs
bjorn: I guess that depends on whether it's actually an error. If there is a 0x0 layer saved somehow, this null container would be the expected return value. You can't create that in Tiled, though (it enforces maps to be 1x1). Might be more useful to raise an error when the dimensions of the returned tile container are not the expected ones?

alexdevteam: Well, if it's not possible to have a 0x0 layer, then I guess it could be interpreted as an error. If this branch is normally inaccessible with any correct Tiled map then maybe it should return `InvalidData` or `Corrupted` or something like that as an error.

bjorn: It's possible to create such a file outside of Tiled and it will load such a file fine. I think it's not useful to consider it an error. It is an error though, if the dimensions of the layer data don't match up with the `width` and `height` attributes. Hence I would suggest a check at a higher level.

bjorn:
Superseded by #128.

PR #60, opened 2019-09-19
Version bump dependencies.

icefoxen:
It doesn't ACTUALLY seem to make any differences, but I'm a little OCD like that.


mattyhall:
Cheers!

PR #61, opened 2019-10-14
Fix tile flipping code

Jengamon:
As @alec-deason rightly pointed out, I messed up the tile flipping code, as it's not encoded into the tileset, but at the layer level. I was kinda dumb, and just assumed that Tile just meant tiles for layers, not tiles for tilesets. To correct my mistake, I rolled back the changes and implemented a new struct: LayerTile. It not only stores the tile gid, but the flags as well. For good measure, I also made it #[derive] Eq, PartialEq, Debug, Clone and Copy, but this is a significant change from the u32 gids.


Jengamon:
Oh, I also added a test case for flipping tiles and stuff.

alec-deason:
Awesome! Glad to see this change. I'll go close my earlier PR since this is a more complete solution.

mattyhall:
Thanks for fixing this :)

PR #64, opened 2019-11-25
Refactor to 2018 edition

TatriX:
Hi! I'd like to add several features, but before that it would be nice to upgrade `rs-tiled` to a recent rust edition. That essentialy is what this PR about.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,0 +1,18 @@
+# Changelog
+All notable changes to this project will be documented in this file.
+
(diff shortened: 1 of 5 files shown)


mattyhall:
Thanks! Is there a reason to bump to 0.9 - I don't see a breaking change? Otherwise happy to merge :)

TatriX:
I just thought that it's a logical step forward, so it's worth of a version bump. Up to you though.

PR #67, opened 2019-12-04
Simplify API by using AsRef<Path>

TatriX:
See commits


bjorn:
@TatriX That looks like a good idea! Would you be interested in rebasing these changes to the latest version?

TatriX:
Hi! Sorry, it's been a while. I don't think I will be able to.

PR #69, opened 2020-02-20
Add "source" value to tiled::Tileset to allow additional functionality.

K4rakara:
I was using this library to load map data for my game, and came across an issue. That issue being that `tiled::Image::source` is a path relative to the path of the tile set it belongs to. As such, trying to load images referenced in an external tile set results in errors. To fix this, I've added a "source" value to `tiled::Tileset` so that one could do
```rust
for tset in map.tilesets.iter() {
    for img in tset.images.iter() {
        let mut image_file = File::open(Path::new(tset.source).join(img.source);
    }
}
```
However, I'm not quite sure what should be used in the case of directly loading a tileset, so one of the 7 tests fail ("" is used as a fallback).


AberDerBart:
I have the same issue, see #94

alexdevteam:
Added in #106.

PR #70, opened 2020-02-24
Make fields on frame pub

mbStavola:
# Context
There doesn't seem to be a way to get at any of the fields on `Frame`.

Let's make it pub so we can poke at it.

# Technical
Ideally we'd add methods that just returned the tile id / duration without allowing modification of the source frame, but I decided to keep this in line with how the rest of the API looks.

Also, a minor rust fmt fix.


mattyhall:
Thanks very much! Fixes #71.

PR #81, opened 2020-05-15
Add support for base64/zstd

tatref:
Hi,

It seems that the default tile layer format is now "Base64 (Zstandard compressed)"
This commits add support for this new compression method.

I tested it on Windows 10 with a map created with the current version of Tiled (2020.05.07)


tatref:
Please don't merge this yet, I'll add some tests in a few days.

PR #82, opened 2020-05-20
add support for base64/zstd

tatref:
Add support for the recent format base64/zstd (default with current version of Tiled).

There is no tests yet, because I am not able to open a map in the assets directory and save it as base64/zstd without changing it's version and various properties in the file.

Maybe we should update all the maps to the newest format? (map version = 1.2)


mattyhall:
Yes, I think it's fine to update the versions of the files! Happy to do this in the next week or so or will accept a PR for it :)

tatref:
I updated all the maps and tilesheets, and saved everything using the recently released Tiled 1.4.

Here is a summary of the changes in the formats:
* map_version: 1.0 -> 1.4
* maps have a Tiled version
* tilesets have a tilecount, and a columns attribute
* layers have an ID
* objectgroups have an ID

mattyhall:
Cheers!

PR #83, opened 2020-05-20
fix deprecated methods in Error

tatref:
`Error::description` and `cause` are deprecated, the new way of using `Error`s (since oct 2018) is with the `source` method.

This removes some warnings when compiling


mattyhall:
Thanks!

PR #84, opened 2020-05-21
add ObjectShape::Point for object layer

tatref:


mattyhall:
Sorry for the delay, and thank you!

PR #86, opened 2020-06-20
Add support for Infinite map

### LINKED ISSUE #76: No support for infinite maps?
Infinite maps are stored as fixed-size chunks in the map file, in the `data` section. When an infinite map is loaded with this library, the `Map` object that results will have the correct width and height, but will only have the first chunk loaded.

Say I have a 32x32 infinite map with only one layer, chunked into 16x16 chunks, and my program attempts to load it with this library. The map will show up as a 32x32 map, but in the map's `layers[0]`, there will be only the first 16x16 chunk that is listed in the map file. No other chunks are loaded.

While an obvious and easy workaround is simply to make the map finite in size, that can be an issue in, for example, the early stages of level design, when you're not even sure how big your map is going to be. Supporting infinite maps would allow the design and testing process to be much simpler.

Are there any plans to support infinite maps in the future?

BlockCat:
This pull request adds support for infinite maps and would resolve #76 

It adds a breaking change to the Layer struct.


mattyhall:
Cool, thank you!

Issue #35, opened 2017-09-21
Add a rendering example


TatriX:
It would be cool to have an opinionated rendering example. I thing piston is a good choice for this.

TatriX:
I'll try to make one and share.

TatriX:
Here it is.
https://github.com/TatriX/rs-tiled-piston-example

bjorn:
@TatriX Would it be a good idea to add this example to this repository?

TatriX:
Well, it seems piston isn't actively maintained. Perhaps `bevy` can be a better choice nowadays.

alexdevteam:
I could do one with the `sfml` crate since that's what i'm most familiar with.

bjorn:
> Well, it seems piston isn't actively maintained. Perhaps `bevy` can be a better choice nowadays.

Ah, alright. For `bevy` there are already [examples](https://github.com/StarArawn/bevy_tiled/tree/main/examples) in `bevy_tiled`.

> I could do one with the `sfml` crate since that's what i'm most familiar with.

Sounds like a good idea.

PR #87, opened 2020-06-23
WIP Add get_tile_rectangle_by_id

tatref:
This is still a work in progress

Helper function for #35

This method return the rectangle where the sprite is located on the image texture.
I tested the math with a few tilesheets (with and without margin/spacings). Seems fine, but I'm not 100% sure!

The rendering code looks like this(using tetra):
```
for layer in map.layers.iter() {
    let tile_width = map.tile_width as f32;
    let tile_height = map.tile_height as f32;
    let texture = ...

    for (j, row) in layer.tiles.iter().enumerate() {
        for (i, tile) in row.iter().enumerate() {
            if tile.gid == 0 {
                continue;
            }
            let p = Vec2::new(i as f32 * tile_width, j as f32 * tile_height);
            let rect = self.map.get_tile_rect_by_id(tile.gid);

            graphics::draw(ctx, texture, DrawParams::default().position(p).clip(rect));
        }
    }
}

```

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -307,6 +307,31 @@ impl Map {
         }
         maximum_ts
     }
+
+    /// Computes the rectangle on the image where the sprite is stored for the given tile ID.
+    /// If the ID is not found in any tileset, or if there is no image associated with the tileset, `None` is returned.
+    /// On success, returns `Some(x, y, w, h)`, where `(x, y)` is the coordinates of the top-left corner, and `(w, h)` are the width and height of the rectangle
+    pub fn get_tile_rectangle_by_id(&self, id: u32) -> Option<(u32, u32, u32, u32)> {
+        let tileset = self.get_tileset_by_gid(id)?;
+        let img = tileset.images.get(0)?; // we suppose there is only 1 image per tileset
+
+        let id = id - 1;
+        let columns =
+            (img.width as u32 - 2 * tileset.margin) / (tileset.spacing + tileset.tile_width);
+
+        // coordinates in tiles
+        let x = id % columns;
+        let y = id.div_euclid(columns);
+
+        // coordinates in pixels
+        let x = tileset.margin + (tileset.tile_width + tileset.spacing) * x;
+        let y = tileset.margin + (tileset.tile_height + tileset.spacing) * y;
+
+        let w = tileset.tile_width;
+        let h = tileset.tile_height;
+
+        Some((x, y, w, h))
+    }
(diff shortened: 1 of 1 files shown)


alexdevteam:
Since this is still a work in progress, I'll label it as a draft.

src/lib.rs
```diff
@@ -307,6 +307,31 @@ impl Map {
...
     }
+
+    /// Computes the rectangle on the image where the sprite is stored for the given tile ID.
+    /// If the ID is not found in any tileset, or if there is no image associated with the tileset, `None` is returned.
+    /// On success, returns `Some(x, y, w, h)`, where `(x, y)` is the coordinates of the top-left corner, and `(w, h)` are the width and height of the rectangle
+    pub fn get_tile_rectangle_by_id(&self, id: u32) -> Option<(u32, u32, u32, u32)> {
+        let tileset = self.get_tileset_by_gid(id)?;
+        let img = tileset.images.get(0)?; // we suppose there is only 1 image per tileset
+
+        let id = id - 1;
```
bjorn, line 318: We need not to subtract 1, we need to subtract `tileset.first_gid`, for this to work with multiple tilesets.

src/lib.rs
```diff
@@ -307,6 +307,31 @@ impl Map {
...
+        let columns =
+            (img.width as u32 - 2 * tileset.margin) / (tileset.spacing + tileset.tile_width);
```
PieKing1215, line 320: This is slightly wrong. Since spacing is between tiles, it doesn't apply to one of the columns (first or last depending on how you think about it)
(ex. tileset with image width = 572, margin = 2, spacing = 4 should be 13 columns, but old method gave ~12.91)
```suggestion
            (img.width as u32 - 2 * tileset.margin - tileset.tile_width) / (tileset.spacing + tileset.tile_width) + 1;
```

bjorn, line 320: That's correct, but rather than subtracting the tile width and adding one, we can instead just add one spacing to the image width.

And actually, the margin is only used on the left and top of the tileset. It does not prevent tiles from getting cut out from the bottom or right sides.

So in the end, this is the calculation used by Tiled:

```suggestion
            (img.width as u32 - tileset.margin + tileset.spacing) / (tileset.tile_width + tileset.spacing);
```

alexdevteam:
There has not been any progress in this PR for more than a year, so I'll close it. If it gets an update I'll reopen it.

PR #89, opened 2020-08-22
Add support for layer offsets

wmiller848:
What
====

* rust-fmt
* Add support for layer offset fields

Example layer with offset

```
<layer id="1" name="Layer 1" width="10" height="10" offsetx="0" offsety="32"> 
```


mattyhall:
Thank you, and apologies for the wait!

PR #91, opened 2020-12-06
Added Terrain/TerrainType so we can now read this info

josebaMdeL:
Added Terrain and TerrainTypes.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -348,6 +348,7 @@ pub struct Tileset {
     /// The Tiled spec says that a tileset can have mutliple images so a `Vec`
     /// is used. Usually you will only use one.
     pub images: Vec<Image>,
+    pub terrain_types: Option<TerrainTypes>,
     pub tiles: Vec<Tile>,
     pub properties: Properties,
 }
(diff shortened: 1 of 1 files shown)


josebaMdeL:
> @josebaMdeL This looks great! Unfortunately, this crate has been reorganized in [c386766](https://github.com/mapeditor/rs-tiled/commit/c3867665183a7165856830f048c2a390f77e88b8) so your changes would need to be rebased. Are you still up for this?
> 
> Another thing you might be aware of, is that terrain information as such is now deprecated and stored as "wang sets" instead since [Tiled 1.5](https://www.mapeditor.org/2021/03/23/tiled-1-5-0-released.html). I'm not sure how interesting it is at this point to still support loading the terrain info. It might be better to just add support for loading Wang sets instead.

Hi!

After a year, I thought this repo was dead. I have been looking to the new tiled versions (I'm still working with Tiled 1.4 on Debian), and seems that a lot of things have changed, the same applies to this crate, so I think the changes I proposed can be ignored. For the moment, I'm working with my version of the crate, but eventually I will need to update it, then I will check if any change is needed.

Anyway, thanks for the answer, and I'm happy to see that this repo has been taken in active hands again. ;)

alexdevteam:
Taking into account the last comment, closing PR until it updates.

PR #92, opened 2021-01-01
Feature-gate zstd so we can target wasm32-unknown-unknown

rparrett:
Another option might be switching to [zstd-rs](https://github.com/KillingSpark/zstd-rs), but I haven't tested this and it seems that there would be a large performance penalty.

People who need compression and wasm can probably recompress with flate, so this seems like a good solution to me.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -12,6 +12,9 @@ edition = "2018"
 
 keywords = ["tiled", "tmx", "map"]
 
+[features]
+default = ["zstd"]
+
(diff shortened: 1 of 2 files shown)


framp:
There is also this approach, using a pure rust zstd: https://github.com/YushiOMOTE/rs-tiled/commit/acf85975f85fcc3e6636e2b0c989431634aa5952

rparrett:
Thanks!

PR #93, opened 2021-02-01
Object visibility needs parsed string, too

dmtaub:
Layer and ObjectGroup are parsing the "visible" string, but Object seems to be missing this expression.
The [TMX docs](https://doc.mapeditor.org/en/stable/reference/tmx-map-format/#object) indicate that visible attribute is expected to be "0" or "1"


mattyhall:
Good catch, thanks!

PR #96, opened 2021-03-07
Added support for Object templates

RossBrunton:
This adds a `template` property to Objects, which is parsed from an
external template file, if its specified using the "template" attribute.

I'm still a little new to Rust, so let me know if the code isn't Rust-y enough.

I had to pass the path around quite a bit, since now Objects need to know the map_path (to load the template, if needed), and objects exist in many places.

I also marked the `x` and `y` properties of objects as optional, since they are listed that way on the docs and templates don't have them.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -279,7 +279,7 @@ impl Map {
                 Ok(())
             },
             "objectgroup" => |attrs| {
-                object_groups.push(ObjectGroup::new(parser, attrs, Some(layer_index))?);
+                object_groups.push(ObjectGroup::new(parser, attrs, Some(layer_index), map_path)?);
                 layer_index += 1;
                 Ok(())
             },
@@ -361,12 +361,13 @@ impl Tileset {
         attrs: Vec<OwnedAttribute>,
         map_path: Option<&Path>,
     ) -> Result<Tileset, TiledError> {
-        Tileset::new_internal(parser, &attrs).or_else(|_| Tileset::new_reference(&attrs, map_path))
+        Tileset::new_internal(parser, &attrs, map_path).or_else(|_| Tileset::new_reference(&attrs, map_path))
     }
 
     fn new_internal<R: Read>(
         parser: &mut EventReader<R>,
         attrs: &Vec<OwnedAttribute>,
+        map_path: Option<&Path>,
     ) -> Result<Tileset, TiledError> {
         let ((spacing, margin, tilecount), (first_gid, name, width, height)) = get_attrs!(
            attrs,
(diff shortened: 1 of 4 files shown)


src/lib.rs
```diff
@@ -896,14 +901,17 @@ pub struct Object {
     pub visible: bool,
     pub shape: ObjectShape,
     pub properties: Properties,
+    pub template: Option<Box<Template>>,
```
mattyhall, line 904: What's the reason for a box here? I don't think the size of a template is great enough for it to be worth doing from a memory perspective

mattyhall:
Apart from above, otherwise looks good. It's a bit sad we have to pass the path around quite so much. Maybe in the future we should think about adding a context struct instead.

bjorn:
@RossBrunton Are you interested in picking up this change again, and rebase it to the latest master?

src/lib.rs
bjorn: Isn't it necessary because of the recursive use of `Object` in `Template`? Can a `struct Object` include an `Option<Object>` as one of its members?

src/lib.rs
```diff
@@ -914,23 +922,29 @@ impl Object {
...
         let id = id.unwrap_or(0u32);
         let gid = gid.unwrap_or(0u32);
+        let x = x.unwrap_or(0f32);
+        let y = y.unwrap_or(0f32);
         let n = n.unwrap_or(String::new());
         let t = t.unwrap_or(String::new());
         let mut shape = None;
         let mut properties = HashMap::new();
+        let template = match tp {
+            Some(path) => Some(Box::new(Template::new(&path, map_path)?)),
```
bjorn, line 945: Since templates could be instantiated a lot of times within the same map, it would make sense to avoid loading the template file repeatedly for each instance. This should be doable by caching the loaded templates using their absolute path.

alexdevteam:
This PR has not been updated for almost a year, so I'll close it. Feel free to reopen if it gets any updates.

RossBrunton:
Aahh! Sorry! Disn't realise you'd looked over it (guess GitHub didn't send me a notification, first I got was that it was closed). I'll rebase it on master, and clean it up.

alexdevteam:
Thank you very much! However, keep in mind that a lot has been refactored in the repo, so the changes may be a bit overwhelming :p

bjorn:
@RossBrunton Great that you'd still like to update this PR! Note that in the meantime, @alexdevteam has introduced a [`ResourceCache`](https://github.com/mapeditor/rs-tiled/blob/master/src/cache.rs) that holds references to loaded tilesets. As I noted before in [a comment here](https://github.com/mapeditor/rs-tiled/pull/96#discussion_r790580315), it would be nice to cache also loaded object templates. Probably this resource cache could be used for that as well.

RossBrunton:
The changes required are looking to be quite major. I'll close this MR and open a new one when I'm done.

PR #98, opened 2021-03-20
Document that map dimensions are in tiles

ISSOtm:
And not, for example, in pixels.


mattyhall:
Looks helpful, thanks!

PR #100, opened 2021-05-07
Added parse_with_file_loader: overrides direct access to the filesystem while loading external tilesets.

koalefant:
Hi, I would like to use external tileset in my game, yet my resources are not in the filesystem, they are packed in an archive. So I am adding a function `parse_with_file_loader(reader, file_loader)` that allows me to supply a closure that is invoked when external tileset is loaded. Do you think this can be upstreamed in any form?

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -4,7 +4,7 @@ use std::collections::HashMap;
 use std::fmt;
 use std::fs::File;
 use std::io::{BufReader, Error, Read};
-use std::path::Path;
+use std::path::{Path, PathBuf};
 use std::str::FromStr;
 use xml::attribute::OwnedAttribute;
 use xml::reader::XmlEvent;
@@ -236,7 +236,7 @@ impl Map {
     fn new<R: Read>(
         parser: &mut EventReader<R>,
         attrs: Vec<OwnedAttribute>,
-        map_path: Option<&Path>,
+        mut external_file_loader: impl FnMut(&str)->Result<Vec<u8>, TiledError>,
     ) -> Result<Map, TiledError> {
         let ((c, infinite), (v, o, w, h, tw, th)) = get_attrs!(
             attrs,
@@ -263,7 +263,7 @@ impl Map {
         let mut layer_index = 0;
         parse_tag!(parser, "map", {
             "tileset" => | attrs| {
-                tilesets.push(Tileset::new(parser, attrs, map_path)?);
+                tilesets.push(Tileset::new(parser, attrs, &mut external_file_loader)?);
                 Ok(())
             },
             "layer" => |attrs| {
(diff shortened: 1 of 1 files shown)


bjorn:
@koalefant This looks very useful! Would you be able to rebase the change to resolve the conflict in `lib.rs`?

Also, @PieKing1215, I see you've already merged this change. Would you be available for providing a code review? I'm looking for additional people to help maintain this crate now that it has moved to the @mapeditor organization.

src/lib.rs
```diff
@@ -236,7 +236,7 @@ impl Map {
     fn new<R: Read>(
         parser: &mut EventReader<R>,
         attrs: Vec<OwnedAttribute>,
-        map_path: Option<&Path>,
+        mut external_file_loader: impl FnMut(&str)->Result<Vec<u8>, TiledError>,
```
alexdevteam, line 239: I'm not sure if this function should return a `TiledError` as `Err`, since it's performing an external data fetch, not something actually related to the Tiled object.

alexdevteam: Also, maybe instead of `Vec<u8>` it should be an object implementing `Read`, that way implementing async later would be easier.

PieKing1215:
> Also, @PieKing1215, I see you've already merged this change. Would you be available for providing a code review? I'm looking for additional people to help maintain this crate now that it has moved to the @mapeditor organization.

Sure, I can review this as well.
I'm using this library pretty extensively in a project so I've liberally merged/fixed several things in my fork, including this PR which I needed for it to work well with ggez.
I have a few of my own small additions that I should really make PRs for as well.

src/lib.rs
```diff
@@ -1290,12 +1287,33 @@ fn parse_impl<R: Read>(reader: R, map_path: Option<&Path>) -> Result<Map, TiledE
     }
 }
 
+fn default_file_loader(map_path: Option<PathBuf>)->impl FnMut(&str)->Result<Vec<u8>, TiledError> {
+    move |source: &str| {
+        let tileset_path = map_path.as_ref()
```
PieKing1215, line 1292: You can do this to avoid needing `to_owned` in the `default_file_loader(Some(path.to_owned()))`s below.
```suggestion
fn default_file_loader(map_path: Option<&Path>)->impl FnMut(&str)->Result<Vec<u8>, TiledError> + '_ {
    move |source: &str| {
        let tileset_path = map_path
```

src/lib.rs
PieKing1215, line 239: Would be nice to extract this into a type alias but I think that would be blocked until either rust-lang/rust#41517 or rust-lang/rust#63063 is stable.

aleokdev:
Closing due to inactivity. Feel free to reopen.

PR #43, opened 2018-04-22
Module reorganization

Delta-Psi:
Splits the monolithic `lib.rs` into several more manageable modules.

PR #104, opened 2021-10-12
Reorganize crate

aleokdev:
This is an updated version of #43. It splits `lib.rs` into the following modules:
- `animation`: For `Frame`
- `error`: For `ParseTileError` & `TiledError`
- `image`: For `Image`
- `layers`: For `LayerTile`, `Layer`, `LayerData`, `ImageLayer` & `Chunk`
- `map`: For `Map` & `Orientation`
- `objects`: For `ObjectGroup`, `ObjectShape` & `Object`
- `properties`: For `Colour`, `PropertyValue` & `Properties`
- `tile`: For `Tile`
- `tileset`: For `Tileset`
- `util`: Private, for all the macros & internal parse functions

While by itself this change won't improve the external interface by much, this reorganization will allow for further development on current types such as `Properties`, which is an alias right now, and is essential to allow for quickly browsing through the library.


mattyhall (approved):
LGTM

bjorn:
@alexdevteam I've just adjusted `bevy_ecs_tilemap` to work with the current git version of `rs-tiled`: https://github.com/bjorn/bevy_ecs_tilemap/commit/eb360e5b2ccc79914cf607ba19adef618b7f2aed

I found the need to prefix every `rs-tiled` type with its specific module name a bit annoying. Is that how things are done in Rust, or is there a way to avoid this?

What I've tried is to add something like `use tiled::map::Map;` at the top, but that didn't really work since it created an ambiguity with `bevy_ecs_tilemap::map::Map`. Personally I would still like the basic types to be available at the top level, like `tiled::Map`, since it allows resolving the ambiguity while avoiding the verbosity of `map::Map`.

alexdevteam:
Yes, I agree that publically exporting essential types on the root is a good idea. Could add a few `pub use`s on `lib.rs`.

PR #106, opened 2021-12-17
Refactor & document

### LINKED ISSUE #94: Add path to Tileset
I want to use multiple Maps with external tile sets. I have two issues:

- I want to manage the tilesets based on their path and avoid loading them multiple times.
- I want to resolve the location of the images used in the tileset, this is not possible, as the paths specified are relative and I don't know from the API, where the external tileset is located.

Both these issues could be solved by adding a `path` field to `Tileset`.

Also, `Image::source` would benefit from being a path that is automatically resolved while parsing.

alexdevteam:
A few miscellaneous improvements and TODO completions:
- Move parse functions to its respective types
- Rename `Colour` for consistency with `PropertyValue::ColorValue`
- Add documentation to map members
- Add `Tileset::source` for obtaining where the tileset actually came from (Closes #94)
- Change `Image::source`'s type to `PathBuf`
- Add `.gitattributes` for line ending consistency
- Add `Tileset::columns`
- `Tile::images` -> `Tile::image`
- `Tileset::images` -> `Tileset::image`
- Update changelog

This PR contains breaking changes.


bjorn:
@alexdevteam This is a big one, I'm afraid it might take until next year for me to get around to reviewing it. Looks good overall though!

alexdevteam:
Merging this if no more reviews are going to be done since this PR blocks other newer changes.

bjorn:
@alexdevteam Some things to consider before you merge:

* Since this crate has been on version 0.9.5 for quite a while and there are some other crates using it, it would be good to provide a small "porting guide". If the changes are small, just updating the `CHANGELOG.md` should be enough. Please do this as part of this change.

* For similar reasons, I would prefer if we'd first do a 0.9.6 patch release with minor (or at least source compatible) changes, before we merge in stuff for 0.10. Is this still possible and does it make sense?

alexdevteam:
> I would prefer if we'd first do a 0.9.6 patch release with minor (or at least source compatible) changes, before we merge in stuff for 0.10. Is this still possible and does it make sense?
We've already done the crate reorganization so I think it would be a hassle to achieve that. In my opinion we should rewrite anything that breaks compatibility before publishing 0.10.0 so users of the crate have to do less work with future updates.

> Since this crate has been on version 0.9.5 for quite a while and there are some other crates using it, it would be good to provide a small "porting guide"
Sure thing, but I think that can be part of another PR as we approach 0.10.

bjorn:
> We've already done the crate reorganization so I think it would be a hassle to achieve that.

Ah, so that reorganization is already a source compatibility break?

> Sure thing, but I think that can be part of another PR as we approach 0.10.

I think it's really better to build this up as we go. It makes it easier to keep track of what has changed since the last release and helps also people who target the master branch.

alexdevteam:
> Ah, so that reorganization is already a source compatibility break?

Yeah, it requires updating all imported items.

> I think it's really better to build this up as we go. It makes it easier to keep track of what has changed since the last release and helps also people who target the master branch.

Alright, I'll add the changelog whenever possible.

src/map.rs
```diff
@@ -1,38 +1,94 @@
...
+                }
+                _ => {}
+            }
+        }
+    }
+
+    /// Parse a file hopefully containing a Tiled map and try to parse it.  If the
+    /// file has an external tileset, the tileset file will be loaded using a path
+    /// relative to the map file's path.
+    pub fn parse_file(path: &Path) -> Result<Self, TiledError> {
```
mattyhall, line 85: I _think_ it's idiomatic to use `AsRef<Path>` rather than `&Path` as it allows you to pass strings in, which is nice.

src/map.rs
```diff
@@ -1,38 +1,94 @@
...
-    util::*,
+    util::{get_attrs, parse_tag},
 };
 
 /// All Tiled files will be parsed into this. Holds all the layers and tilesets
 #[derive(Debug, PartialEq, Clone)]
 pub struct Map {
+    /// The TMX format version this map was saved to.
     pub version: String,
+    /// The orientation of this map.
```
mattyhall, line 26: A bit of a nit (and very opinionated so worth someone else's opinion) but not sure comments like this add much

mattyhall (commented):
Overall filling out the docs, cleaning up the interface a bit and fixing Image is good. I'd agree that some caution is needed in terms of rolling this out - there are quite big changes like renaming fields

src/map.rs
alexdevteam: You're right, there was another PR just for this but it's outdated so I can implement it here as well.

src/map.rs
alexdevteam: I agree, I just have the tendency to always add docs to everything.

src/map.rs
```diff
@@ -97,13 +149,14 @@ impl Map {
...
-            background_colour: c,
+            background_color: c,
             infinite: infinite.unwrap_or(false),
+            source: map_path.and_then(|p| Some(p.to_owned())),
         })
     }
 
     /// This function will return the correct Tileset given a GID.
-    pub fn get_tileset_by_gid(&self, gid: u32) -> Option<&Tileset> {
+    pub fn tileset_by_gid(&self, gid: u32) -> Option<&Tileset> {
```
bjorn, line 159: This rename should be reflected in the `README.md` as well.

src/tileset.rs
```diff
@@ -67,6 +94,18 @@ impl Tileset {
             },
         });
 
+        let columns = match columns {
+            Some(col) => col,
+            None => match &image {
+                None => {
+                    return Err(TiledError::MalformedAttributes(
+                        "No <image> and no <columns> in <tileset>".to_string(),
```
bjorn, line 102: Message is a bit confusing because `columns` is an attribute rather than an element.

tests/lib.rs
```diff
@@ -1,26 +1,21 @@
 use std::fs::File;
 use std::path::Path;
 use tiled::{
-    error::TiledError, layers::LayerData, map::Map, parse, parse_file, parse_tileset,
-    properties::PropertyValue,
+    error::TiledError, layers::LayerData, map::Map, properties::PropertyValue, tileset::Tileset,
 };
 
-fn read_from_file(p: &Path) -> Result<Map, TiledError> {
+fn parse_map_without_source(p: impl AsRef<Path>) -> Result<Map, TiledError> {
```
bjorn, line 7: What does `parse_map_without_source` really mean?

src/lib.rs
```diff
@@ -8,55 +8,3 @@ pub mod properties;
...
-/// relative to the map file's path.
-pub fn parse_file(path: &Path) -> Result<Map, TiledError> {
-    let file = File::open(path)
-        .map_err(|_| TiledError::Other(format!("Map file not found: {:?}", path)))?;
-    parse_impl(file, Some(path))
-}
-
-/// Parse a buffer hopefully containing the contents of a Tiled file and try to
-/// parse it.
-pub fn parse<R: Read>(reader: R) -> Result<Map, TiledError> {
```
bjorn, line 51: Move of `parse` needs to be reflected in the `README.md`.

src/map.rs
alexdevteam: I didn't change the README at all, all changes are in the changelog.

src/lib.rs
alexdevteam: It's in the changelog. I could add a notice for old users pointing to the CHANGELOG file.

tests/lib.rs
alexdevteam: Parses a map from the given path, without setting the `source` attribute in the Map struct.

src/lib.rs
bjorn: I just meant, the `README.md` needs to be updated, because it contains an example that still uses the old API.

tests/lib.rs
bjorn: When is that useful?

alexdevteam: Nowhere, you're right, I just did a blind translation from the old function without thinking too much.

alexdevteam: Had to look at it from my computer; it is useful because tests check if maps with the same data but from different sources are equal. The `source` property of them needs to always be set to `None`, otherwise all equality checks will fail.

bjorn: Hmm, if that's really the only difference I would suggest to do that only in the tests.

But, in general I don't really understand the need for the `source` on `Map`. Wouldn't it be better, to resolve the paths to absolute paths during loading, so that after loading we no longer need to know from where a map was loaded?

.gitattributes
```diff
@@ -4,6 +4,7 @@
 # Explicitly declare text files we want to always be normalized and converted
 # to native line endings on checkout.
 *.rs text
+*.tmx text eol=crlf
```
bjorn, line 7: Really needs a comment as to why this is needed.

tests/lib.rs
alexdevteam: Hmm, good idea. We could move `source` to `Tileset` instead and use absolute paths there.

src/tileset.rs
```diff
@@ -67,6 +94,18 @@ impl Tileset {
...
 
+        let columns = match columns {
+            Some(col) => col,
+            None => match &image {
+                None => {
+                    return Err(TiledError::MalformedAttributes(
+                        "No <image> nor columns attribute in <tileset>".to_string(),
+                    ))
+                }
+                Some(image) => image.width as u32 / width,
```
bjorn, line 105: I guess this fallback should take spacing and margin into account? So `(image.width - margin + spacing) / (tile_width + spacing)`.

src/tileset.rs
```diff
@@ -153,33 +194,46 @@ impl Tileset {
...
+                Ok(())
+            },
         });
 
+        let columns = match columns {
+            Some(col) => col,
+            None => match &image {
+                None => {
+                    return Err(TiledError::MalformedAttributes(
+                        "No <image> and no <columns> in <tileset>".to_string(),
```
bjorn, line 220: Duplicate code here.

alexdevteam:
Sorry for the size of this PR. Hopefully this should be the last enormous one for a while. There are a few more things I'd like to change in the interface, such as GIDs getting their own type (Spoilers for next PR), but hopefully those changes should all be much more localized and easier to identify.

bjorn (approved):
Looks good to me!

PR #110, opened 2021-12-26
Add support for multiline string properties

PieKing1215:
If you enter a multiline string into a string property in the editor, it saves the string inside the xml node instead of inside the "value" attribute:
```xml
<property name="prop3">Line 1
Line 2
Line 3,
  etc
   </property>
```
which should be parsed as:
```
"Line 1\r\nLine 2\r\nLine 3,\r\n  etc\r\n   "
```

(I believe Tiled saves the file with the system's native line endings right? Not sure if the parsing in this PR should normalize it)

This PR adds support for parsing multiline strings like this.
It also adds such a property to the test files and adds a test to check that it works.

(feel free to suggest changes)


alexdevteam (approved):
Looks good!

bjorn:
> (I believe Tiled saves the file with the system's native line endings right? Not sure if the parsing in this PR should normalize it)

Doesn't that happen automatically depending on whether the file is opened in text mode? That's what happens in Tiled at least, but it might be as a result of using Qt.

alexdevteam:
> Not sure if the parsing in this PR should normalize it
Didn't notice this, it'd be better if you could normalize line endings from code

PR #111, opened 2021-12-26
Add layer id parsing

PieKing1215:
Adds support for parsing the layer "id" attribute.
Unlike layer index, the id of a layer always stays the same, even if the order is changed or other layers are added/deleted.
(eg. this can be useful for identifying a layer in an int property on an object)

### DIFF
--- a/src/layers.rs
+++ b/src/layers.rs
@@ -75,7 +75,7 @@ impl Layer {
             ],
             required: [
                 ("name", name, |v| Some(v)),
-                ("id", id, |v:String| v.parse::<u32>().ok()),
+                ("id", id, |v:String| v.parse().ok()),
             ],
             TiledError::MalformedAttributes("layer must have a name".to_string())
         );
--- a/src/layers.rs
+++ b/src/layers.rs
@@ -54,6 +54,7 @@ pub struct Layer {
     pub tiles: LayerData,
     pub properties: Properties,
     pub layer_index: u32,
+    pub id: u32,
 }
 
 impl Layer {
@@ -64,7 +65,7 @@ impl Layer {
         layer_index: u32,
         infinite: bool,
     ) -> Result<Layer, TiledError> {
-        let ((o, v, ox, oy), n) = get_attrs!(
+        let ((o, v, ox, oy), (n, id)) = get_attrs!(
             attrs,
             optionals: [
                 ("opacity", opacity, |v:String| v.parse().ok()),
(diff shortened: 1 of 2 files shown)


alexdevteam (changes requested):
Could you add docs for the `id` members?

alexdevteam (approved):

src/layers.rs
```diff
@@ -74,6 +77,7 @@ impl Layer {
             ],
             required: [
                 ("name", name, |v| Some(v)),
+                ("id", id, |v:String| v.parse().ok()),
             ],
             TiledError::MalformedAttributes("layer must have a name".to_string())
```
bjorn, line 82: Seems like we need to either update this string or don't require the `id` attribute. I don't personally see why the IDs should be required, so I'd suggest moving it to `optionals`.

src/objects.rs
```diff
@@ -28,15 +31,17 @@ impl ObjectGroup {
...
                 ("opacity", opacity, |v:String| v.parse().ok()),
                 ("visible", visible, |v:String| v.parse().ok().map(|x:i32| x == 1)),
                 ("color", colour, |v:String| v.parse().ok()),
                 ("name", name, |v:String| v.into()),
             ],
-            required: [],
+            required: [
+                ("id", id, |v:String| v.parse().ok()),
+            ],
             TiledError::MalformedAttributes("object groups must have a name".to_string())
```
bjorn, line 45: If they must have a name, should `name` be moved to `required`? I don't see why they should have a name though, so maybe the message should be fixed.

bjorn (commented):
Just leaving two comments about tweaks we may want to make.

alexdevteam:
@PieKing1215 Could you make another PR with the requested changes?

src/layers.rs
PieKing1215: I was under the impression that layers are required to have the `id` attribute since if you edit the file to remove it (or set it to something invalid), Tiled adds one back.

bjorn: Hmm, yeah layers will always get an ID assigned, but there is just no reason to fail loading a map just because there is no `id` attribute.

src/objects.rs
PieKing1215: Yeah I think that message should change, you can get a layer to have no `name` attribute just by setting it to empty in Tiled.
Not sure why it's optional here but required in `Layer`, they should both be optional I think.

src/layers.rs
PieKing1215: Fair. How should that be handled though, store `id` as `Option<u32>` or keep it raw and assign a default value?
We could use 0 since it's normally unused, or we could base it on `nextlayerid` like Tiled (it might be weird to have hidden functionality like that though)

bjorn: I think it should stay on 0. Indeed layers would never get ID 0, so it's a suitable value for when the attribute isn't there, and indeed automatically assigning IDs isn't useful outside of an editor.

PR #119, opened 2021-12-29
Make layer ids optional and make tile layer name optional (like object layers)

PieKing1215:
Follow-up for #111

Layer id defaults to 0 if missing, which is normally unused.
Tile layer name defaults to an empty String if missing, same as object layers.

### DIFF
--- a/src/layers.rs
+++ b/src/layers.rs
@@ -127,6 +127,9 @@ pub struct ImageLayer {
     pub image: Option<Image>,
     pub properties: Properties,
     pub layer_index: u32,
+    /// The ID of the layer, as shown in the editor.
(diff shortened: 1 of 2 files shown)


alexdevteam (approved):

bjorn (commented):
Looks good! Please just also synchronize the `ImageLayer` while you're at it, adding `id` and making `name` optional.

alexdevteam (commented):
Maybe `name` should be `Option<String>` then? Otherwise looks good.

PieKing1215:
I would lean towards a missing name meaning empty string since that's how Tiled represents it, but I could change it to an `Option` if you really want (would make this breaking).

alexdevteam:
It's alright, the no-name case is not common enough to consider. Merging then.

Issue #116, opened 2021-12-28
CHANGELOG.md outdated


perlindgren:
Just observed that the `CHANGELOG.md` is lagging behind (0.9.2 vs 0.9.5), not sure what has changed though.

/Per

PR #121, opened 2021-12-29
Start getting changelog up to date

PieKing1215:
Started work on #116

For 0.9.1-0.9.5, I used the dates of crates.io releases. I wasn't sure whether to use that or the bump commit dates, as in some cases they are quite different.
0.9.0 doesn't have a crates.io release, so it uses the commit date.

I definitely could have missed some things, especially for 0.9.0 since the repo seems to have jumped directly from 0.8.0 to 0.9.0 while on crates.io there was a 0.8.1 as well, not sure where the cutoff was.


alexdevteam:
All unreleased changes have been written in #106, just so you know.

PieKing1215:
Oh cool I didn't see that, your unreleased section is much more thorough so if/when these get merged together make sure to keep yours.

CHANGELOG.md
```diff
@@ -10,18 +10,60 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 
 - Support for 'object'-type properties.
+- Support for multi-line string properties.
+- Layer id parsing.
 
-## [0.9.2] - 2020-Apr-25
+### Changed
+
```
alexdevteam, line 17: I'd prefer if these spaces before the bullet lists are removed like in https://keepachangelog.com/en/1.0.0/, so that we can fit more data in less visual space.

bjorn (approved):
Looks fine to me, but because of it being a bit of a historical dig, I guess it would make sense to see if @mattyhall has anything to comment.

alexdevteam (approved):

bjorn:
Thanks @PieKing1215!

Issue #107, opened 2021-12-23
README is outdated


aleokdev:
The README is outdated:
- It refers to `0.9.3` as the latest version.
- The http://blog-buch.rhcloud.com/ link is currently dead.
- The build badge is invalid now (There's no longer CI set up).
- The "Things missing" list could be updated.

alexdevteam:
It still is, and requires a few updates.

PR #122, opened 2021-12-29
Start updating README.md

PieKing1215:
Start work on #107

Updated listed version from 0.9.3 to 0.9.5.
Updated CI link to GH actions build.
Remove things from "Things missing" that were added.
Update example to use `parse_file` instead of manually turning the `File` into `BufReader` and using `parse` (note: this is also updated again in #106)

This PR doesn't add anything new to "Things missing" yet, and doesn't fix the bottom link.


bjorn:
> This PR doesn't add anything new to "Things missing" yet, and doesn't fix the bottom link.

I think maybe we should juts remove that section. I personally see little point in trying to document all the things that are missing in the `README.md`, we could instead open issues for things we know are missing.

Regarding the bottom link, I think it could link to https://opengameart.org/content/sci-fi-interior-tiles instead.

bjorn (approved):

alexdevteam (approved):
Looks great!

PR #125, opened 2022-01-08
Layer.get_tile(x, y) implemented.

singalen:
I needed this function. Maybe other users process layers in chunks, but I still view this function as useful.

### DIFF
--- a/src/layers.rs
+++ b/src/layers.rs
@@ -57,6 +57,15 @@ pub struct Layer {
     /// The ID of the layer, as shown in the editor.
     /// Layer ID stays the same even if layers are reordered or modified in the editor.
     pub id: u32,
+
+    /// Derived from the first chunk. There's no good way to generally find out
+    /// the chunk size, even though
+    /// [editorsettings doc](https://doc.mapeditor.org/en/stable/reference/tmx-map-format/#chunksize)
+    /// suggests it's the same for the entire map.
+    /// Maybe in future Tiled will support different chunk sizes in one map, but
+    /// I doubt it.
+    pub chunk_width: u32,
+    pub chunk_height: u32,
 }
 
 impl Layer {
@@ -98,6 +107,16 @@ impl Layer {
             },
         });
 
+        let mut chunk_width = 0;
+        let mut chunk_height = 0;
+
+        if let LayerData::Infinite(chunks) = &tiles {
+            if let Some((_i, chunk)) = chunks.iter().next() {
+                chunk_width = chunk.width;
+                chunk_height = chunk.height;
+            }
+        }
+
         Ok(Layer {
             name: n.unwrap_or(String::new()),
             opacity: o.unwrap_or(1.0),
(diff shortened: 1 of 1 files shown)


src/layers.rs
```diff
@@ -57,6 +57,15 @@ pub struct Layer {
...
     pub id: u32,
+
+    /// Derived from the first chunk. There's no good way to generally find out
+    /// the chunk size, even though
+    /// [editorsettings doc](https://doc.mapeditor.org/en/stable/reference/tmx-map-format/#chunksize)
+    /// suggests it's the same for the entire map.
+    /// Maybe in future Tiled will support different chunk sizes in one map, but
+    /// I doubt it.
+    pub chunk_width: u32,
+    pub chunk_height: u32,
```
bjorn, line 68: At the moment Tiled only supports a fixed chunk size, where the size is actually stored on the map, but as you noticed this is an editor setting. I could imagine to add support for dynamic chunk sizes in the future, but it would likely be only added on demand. While that might be unlikely, I'm not so keen on adding these members.

bjorn (commented):
The way I see it, the chunks are just an artifact of the storage format. The in-memory representation does not need to match it. Tiled itself always stores tile layers using 16x16 chunks, and this is unlikely to become configurable, because it's just an optimization over using a basic `map<point, tile>` kind of storage, which is otherwise irrelevant to the user.

Similarly, `rs-tiled` could choose some static chunk size as well (possibly based on profiling, which I neglected to do so far) and just load the saved chunks into its own static chunks.

Or we could provide a setting for the chunk size used by `rs-tiled`, since that might be helpful for tile layer rendering code that also likes to render in chunks. But this adds complexity and I'm not sure if it's worth it.

Could you imagine to make a patch for using a chunk size independent of the one used in the stored file?

src/layers.rs
singalen: Currently, in order to retrieve a chunk, the code needs to know its top-left corner. E.g. in order to retrieve tile at (17, 17) from `Infinite(HashMap<(i32, i32), ...)`, code needs to deduce the (16,16) key; trying (17,17) will not find it.

This is why I kinda needed these fields.

Ideal general solution would be to use a quad-tree to map coordinates to chunks. But as the values don't really change, I thought a simpler solution would work fine for many years. Quad-tree also has a logarithmic access time, and direct indexing is constant.

Another solution is to retrieve a (0,0) chunk on each `get_tile(x, y)` call, read ITS size, assume that all chunks have this size, and then evaluate index of the needed chunk. But it's still one extra operation where I don't want it, namely in a rendering code.

Here are the options I see. What do you say?

bjorn: I'd prefer the hashmap to a quad tree, indeed because lookups should be very fast. Alternatively the hashmap could even be an array of chunk references, which could waste some memory but avoids the hashing. The second solution is definitely also something we'd want to avoid.

But, could you share your thoughts on the suggestion to use a fixed chunk size in `rs-tiled`, independent from the chunks found in the TMX file, and while loading the chunks in the TMX file, to just copy over the tile data in the fixed chunks data structure? This is the way Tiled works internally.

bjorn:
@singalen Could you rebase this change, now that #128 has been merged?

Regarding the `chunk_width/height` topic, if you don't want to try the fixed chunk sizes change I suggested, I'm fine with merging this with these attributes in place for now. We can revisit that topic in a future change. Maybe we could at least move them to be part of `LayerData::Infinite`, so that we don't have them sit around for finite maps?

alexdevteam:
Superceded by #157

PR #126, opened 2022-01-08
Any interest in EditorSettings?

singalen:
I honestly see no real point, but maybe you want it for functional completeness. If not, I'll close this PR.

I needed to read the chunk size, but I don't think EditorSettings is an authoritative source.

### DIFF
--- a/src/map.rs
+++ b/src/map.rs
@@ -36,6 +36,10 @@ pub struct Map {
     /// The background color of this map, if any.
     pub background_color: Option<Color>,
     pub infinite: bool,
+    /// "various editor-specific settings, which are generally not relevant
+    /// when reading a map" the doc says, but the only other way to read
+    /// an infinite map's chunk size is directly from chunks.
+    pub editor_settings: EditorSettings
 }
 
 impl Map {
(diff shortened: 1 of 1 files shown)


bjorn:
> I needed to read the chunk size, but I don't think EditorSettings is an authoritative source.

Yeah, at the moment it is, but it's not very nice to rely on this. Personally I think `rs-tiled` should use a chunk size that is independent of the stored files. See also https://github.com/mapeditor/rs-tiled/pull/125#pullrequestreview-847580483.

bjorn:
I guess alternatively, if we recognize the idea of dynamic chunk sizes is just a bad idea, then the implication is that storing the size on each chunk was unnecessary, and we could introduce attributes outside of the `editorsettings` that store the size and get rid of the `width` and `height` attributes on each `chunk` element. I think we should then also store the chunk coordinates in chunks rather than tiles. The main problem with this, of course, is the breaking of compatibility.

bjorn:
Stored chunk sizes are no longer relevant now that #157 was merged. Editor settings might be relevant in the future, but likely only once we would start work on #171, so closing this for now.

PR #128, opened 2022-01-24
Flatten Vec<Vec<LayerTile>> to Vec<LayerTile>

### LINKED ISSUE #54: Vec<Vec<T>> is inefficient
It would be better to index a single large Vec as a 2D object, Row Order preferably (As this is the convention used by C and most other languages)

https://stackoverflow.com/a/2151141

Why this is inefficient:
Simple, really: Heap fragmentation. This allocates one new vector for each column of the vector. I.e. in the case of `Layer`, a large layer, say, 1024x1024, will have 1025 length 1024 vectors allocated. A single vector is nicer to the kernel (One big allocation vs multiple small ones), and for iterating over the whole map, it performs quite a bit better, as the CPU can easily predict the access pattern on it's own without assistance.

bjorn:
This simplifies code and should be better on memory usage since it is less fragmented.

Also added `width` and `height` attributes to `Layer`.

Closes #54

### DIFF
--- a/tests/lib.rs
+++ b/tests/lib.rs
@@ -22,14 +22,12 @@ fn test_gzip_and_zlib_encoded_and_raw_are_the_same() {
     assert_eq!(z, zstd);
 
     if let LayerData::Finite(tiles) = &c.layers[0].tiles {
-        assert_eq!(tiles.len(), 100);
-        assert_eq!(tiles[0].len(), 100);
-        assert_eq!(tiles[99].len(), 100);
-        assert_eq!(tiles[0][0].gid, 35);
-        assert_eq!(tiles[1][0].gid, 17);
-        assert_eq!(tiles[2][0].gid, 0);
-        assert_eq!(tiles[2][1].gid, 17);
-        assert!(tiles[99].iter().map(|t| t.gid).all(|g| g == 0));
+        assert_eq!(tiles.len(), 100 * 100);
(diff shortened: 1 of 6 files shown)


tests/lib.rs
```diff
@@ -22,14 +22,12 @@ fn test_gzip_and_zlib_encoded_and_raw_are_the_same() {
...
-        assert_eq!(tiles[1][0].gid, 17);
-        assert_eq!(tiles[2][0].gid, 0);
-        assert_eq!(tiles[2][1].gid, 17);
-        assert!(tiles[99].iter().map(|t| t.gid).all(|g| g == 0));
+        assert_eq!(tiles.len(), 100 * 100);
+        assert_eq!(tiles[0].gid, 35);
+        assert_eq!(tiles[100].gid, 17);
+        assert_eq!(tiles[200].gid, 0);
+        assert_eq!(tiles[200 + 1].gid, 17);
+        assert!(tiles[9900..9999].iter().map(|t| t.gid).all(|g| g == 0));
```
bjorn, line 30: This is a bit hard to read, but it should be fixed soon when we add `Layer.get_tile` (#125).

alexdevteam (approved):
Looks good!

alexdevteam:
Could you add any interface changes to CHANGELOG.md?

alexdevteam (approved):

PR #129, opened 2022-01-24
CI: Added caching of Rust stuff to potentially speed up action

bjorn:


bjorn:
Seems like this shaves a significant amount off the build time. :-)

alexdevteam (approved):

PR #131, opened 2022-01-25
Contain all layer types in an enum

### LINKED ISSUE #130: Have layer common data in a single struct
Pack all common data from all types of layers in a single struct and specialize between different layer types (e.g. group, image, tile) with an enum. This will allow to have all map layers in a single member and remove the `layer_index` member.

alexdevteam:
Closes #130 & supersedes #53.
Fits all layer types into a single `LayerType` enum and places all common data (ID, name, properties, etc) into a struct.


src/tile.rs
```diff
@@ -16,7 +16,7 @@ pub struct Tile {
     pub id: u32,
     pub image: Option<Image>,
     pub properties: Properties,
-    pub objectgroup: Option<ObjectGroup>,
+    pub objectgroup: Option<(ObjectLayer, Properties)>,
```
alexdevteam, line 19: Not sure about this: Tile may contain an object group, and this member doesn't include things such as name or ID (Those are stored in `Layer`). If we used `Layer` then the usage of `LayerType::ObjectLayer` would be implicit and it would create invalid states (e.g. tile has a tile layer in it).

tests/lib.rs
```diff
@@ -21,15 +22,19 @@ fn test_gzip_and_zlib_encoded_and_raw_are_the_same() {
...
     assert_eq!(z, zstd);
 
-    if let LayerData::Finite(tiles) = &c.layers[0].tiles {
-        assert_eq!(tiles.len(), 100 * 100);
-        assert_eq!(tiles[0].gid, 35);
-        assert_eq!(tiles[100].gid, 17);
-        assert_eq!(tiles[200].gid, 0);
-        assert_eq!(tiles[200 + 1].gid, 17);
-        assert!(tiles[9900..9999].iter().map(|t| t.gid).all(|g| g == 0));
+    if let LayerType::TileLayer(layer) = &c.layers[0].ty {
```
alexdevteam, line 25: Doing if-lets in this way is pretty verbose, I might implement `TryInto<XLayer>` for `LayerType`.

src/tile.rs
alexdevteam: If I am correct, this object group refers to the tile collisions, so no data that is in the `Layer` struct actually matter here, not even the properties (at least in 99% in the cases) since it's inaccessible from the editor anyways.

bjorn: Hmm, indeed even the properties are not accessible in the editor as far as I know, so potentially just `Option<ObjectGroup>` would suffice.

src/layers.rs
```diff
@@ -42,42 +43,145 @@ impl LayerTile {
...
-    /// The tiles are arranged in rows. Each tile is a number which can be used
-    ///  to find which tileset it belongs to and can then be rendered.
-    pub tiles: LayerData,
+    pub opacity: f32,
     pub properties: Properties,
-    pub layer_index: u32,
-    /// The ID of the layer, as shown in the editor.
-    /// Layer ID stays the same even if layers are reordered or modified in the editor.
-    pub id: u32,
+    pub ty: LayerType,
```
bjorn, line 111: Let's just call this `layer_type`.

src/layers.rs
```diff
@@ -42,42 +43,145 @@ impl LayerTile {
...
+        let ((opacity, visible, offset_x, offset_y, name), id) = get_attrs!(
             attrs,
             optionals: [
                 ("opacity", opacity, |v:String| v.parse().ok()),
                 ("visible", visible, |v:String| v.parse().ok().map(|x:i32| x == 1)),
                 ("offsetx", offset_x, |v:String| v.parse().ok()),
                 ("offsety", offset_y, |v:String| v.parse().ok()),
                 ("name", name, |v| Some(v)),
+            ],
+            required: [
```
bjorn, line 131: We just made the `id` not required in aa8214abe2e33dc9988e1255eb7dceab604d9447. Even if Tiled will always write it, I don't see a reason to require it. Tiled itself for example needs to be able to load maps without these, for backwards compatibility.

src/layers.rs
```diff
@@ -42,42 +43,145 @@ impl LayerTile {
...
-#[derive(Debug, PartialEq, Clone)]
+#[derive(Clone, PartialEq, Debug)]
+pub enum LayerType {
+    TileLayer(TileLayer),
+    ObjectLayer(ObjectLayer),
+    ImageLayer(ImageLayer),
+    // TODO: Support group layers
+}
+
+impl LayerType {
```
bjorn, line 54: I'm not entirely convinced we need this. It seems to me that it encourages use of `if` and `unwrap`, where I would expect in most cases it's best to do a `match`.

bjorn (commented):
Great that you went ahead with this change, and I think it turned out pretty nice!

It's also interesting to see how you can juggle those properties around, to make them part of `Layer` instead of each specific layer type.

Btw, why is it still marked as draft, even though it appears to compile and test out fine?

src/layers.rs
alexdevteam: I was considering a name like that, but it is way too verbose (plus the `layer` part is already implicit from the parent type). It cannot be named `type` either because it's a keyword, so think `ty` works fine. In my mind `LayerType` is a type-specific "extension" of `Layer` so it makes sense to have a short name for it.

src/layers.rs
alexdevteam: Added these because of [the examples](https://github.com/mapeditor/rs-tiled/pull/131#discussion_r792079569), where I thought that matching turned out way too verbose. I can revert the changes if you want.

alexdevteam:
> Btw, why is it still marked as draft, even though it appears to compile and test out fine?

Just because I wasn't 100% sure of the final interface. Group layers will get their own PR, since they weren't implemented even before these changes.

src/layers.rs
bjorn: We can leave that for now. It does appear useful there, but that is testing code, not example.

src/layers.rs
bjorn: I think it's really unfortunate that Rust devs decided to make "type" a keyword...

Anyway, I'd still prefer a verbose name over `ty`. I'm using the [same name in Tiled](https://github.com/mapeditor/tiled/blob/master/src/libtiled/layer.h#L85), so it seems rather natural to me.

tests/lib.rs
```diff
@@ -21,15 +22,15 @@ fn test_gzip_and_zlib_encoded_and_raw_are_the_same() {
     assert_eq!(z, c);
     assert_eq!(z, zstd);
 
-    if let LayerData::Finite(tiles) = &c.layers[0].tiles {
+    if let LayerData::Finite(tiles) = &c.layers[0].layer_type.as_tile_layer().unwrap().tiles {
```
bjorn, line 25: Hmm, what if we would put `as_tile_layer` on the `Layer`, so that we could get `&c.layers[0].as_tile_layer()`?

bjorn (dismissed):

bjorn:
+218 −168... always the same, if you clean and remove lots of duplicate code, you still usually end up with more code! :S

tests/lib.rs
alexdevteam: Oops, saw this too late

alexdevteam: Imo not a good decision because `as_tile_layer` only borrows `LayerType`, not the entire `Layer`

bjorn: Ah, this type of consideration is still new to me. :-)

bjorn (dismissed):

alexdevteam:
Okay, I swear this is the last time I ask you to re-review :p

bjorn (approved):

PR #132, opened 2022-01-25
Expose types at the top level

bjorn:
While the modules are great for organizing the code internally, it is more convenient for users of the create when the types are available at the top level. This avoids the verbosity like `map::Map`.

As discussed at https://github.com/mapeditor/rs-tiled/pull/104#issuecomment-1020333980.


bjorn:
If this is fine I think we should adjust the changelog as well, since it no longer needs to mention all those moves.

alexdevteam (commented):
Alright, if you update the changelog it's good to go.

CHANGELOG.md
```diff
@@ -11,37 +11,21 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - `Tileset::columns`.
 - `Layer::id`, `Layer::width` and `Layer::height`.
 - Support for 'object'-type properties.
-- Documentation for map members.
-- Tests for `tiled_base64_zstandard.tmx`.
-- `.gitattributes` for line ending consistency.
```
alexdevteam, line 16: Why'd you remove these additions?

bjorn: I think the addition of documentation or especially `.gitattributes` or fixing of markdown formatting is not noteworthy for people who want to know what changed in rs-tiled, so I would not put them in the changelog. If people want to know all the details, they could just look at the git history.

But I'll put them back for now, since their removal was unrelated to this change. Let me know your opinion though, since I'd still like to remove these.

alexdevteam (approved):

bjorn:
@alexdevteam For future changes, please only approve. There is no need to merge because I can do this, so in this case I'd rather decide when to merge and how to format the final commit message.

PR #133, opened 2022-01-25
Add SFML example

### LINKED ISSUE #35: Add a rendering example

alexdevteam:
Adds a rough demo using SFML for rendering. Closes #35.

Extremely WIP. Right now the example is really long because a lot of boilerplate is required. If you realize any way to reduce this boilerplate without making the example more specific, let me know.


examples/sfml/tilesheet.rs
```diff
@@ -0,0 +1,73 @@
...
+
+    pub fn texture(&self) -> &Texture {
+        &self.texture
+    }
+
+    pub fn tile_rect(&self, gid: u32) -> Option<IntRect> {
+        if gid == 0 {
+            return None;
+        }
+        let id = gid - self.tileset.first_gid;
```
bjorn, line 37: Since this is the `Tilesheet` class, I think it would be better if a local tile index was passed in. We should avoid proliferating the `gid` to this level. Then this function shouldn't need to return `Option` either.

examples/sfml/tilesheet.rs
```diff
@@ -0,0 +1,73 @@
...
+        };
+
+        Tilesheet { texture, tileset }
+    }
+
+    pub fn texture(&self) -> &Texture {
+        &self.texture
+    }
+
+    pub fn tile_rect(&self, gid: u32) -> Option<IntRect> {
```
bjorn, line 33: Since we're building an SFML example here, wouldn't it be better to just directly return a `FloatRect`, so we can remove the separate `tile_uv` function?

examples/sfml/mesh.rs
```diff
@@ -0,0 +1,47 @@
...
+    /// Create a new mesh with capacity for the given amount of vertices.
+    pub fn with_capacity(vertices: usize) -> Self {
+        Self(Vec::with_capacity(vertices))
+    }
+
+    /// Add a quad made up of vertices to the mesh.
+    pub fn add_quad(&mut self, position: Vector2f, size: f32, uv: FloatRect) {
+        self.0.push(Vertex::new(
+            position,
+            Color::WHITE,
```
bjorn, line 18: Since all the vertices are white, let's use `Vertex::with_pos_coords` instead.

examples/sfml/mesh.rs
```diff
@@ -0,0 +1,47 @@
+use sfml::{
+    graphics::{Color, Drawable, FloatRect, PrimitiveType, Vertex},
+    system::Vector2f,
+};
+
+pub struct VertexMesh(Vec<Vertex>);
+
+impl VertexMesh {
+    /// Create a new mesh with capacity for the given amount of vertices.
+    pub fn with_capacity(vertices: usize) -> Self {
```
bjorn, line 10: Maybe it makes sense to rename this struct to `QuadsMesh`, given that it can only be used to render quads? Then it would take `quads` argument and the `* 4` for the capacity can be moved in here.

bjorn (commented):
Works fine! Though we really need a nicer example map I guess. :-)

I've left a number of comments, most of which I'll also address by pushing my changes to the example, but let me know if you disagree about any of them.

It took me a bit to figure out the dependencies to install. In the end I had to install `libcsfml-dev`, `libsfml-dev`, `build-essential` (not entirely sure) and `clang` before it would work. Probably something we need to document somewhere.

bjorn:
> I've left a number of comments, most of which I'll also address by pushing my changes to the example, but let me know if you disagree about any of them.

Hmm, so I've already addressed most of my comments, since I anyway wanted to make sure the changes made sense, given that I'm still learning Rust. However, it appears I can't push to your branch. I've pushed the change here instead: https://github.com/bjorn/rs-tiled/commit/b547ffa8a74479f1927419fdf083cca24a8affc1

alexdevteam:
> It took me a bit to figure out the dependencies to install. In the end I had to install `libcsfml-dev`, `libsfml-dev`, `build-essential` (not entirely sure) and `clang` before it would work. Probably something we need to document somewhere.

The sfml crate is supposed to vendor its own version of sfml, but for some reason still requires those packages as dependencies. I've had issues with that as well in the past, I'll create an issue in the sfml-rs repo.

examples/sfml/tilesheet.rs
```diff
@@ -0,0 +1,73 @@
...
+
+impl Tilesheet {
+    /// Create a tilesheet from a Tiled tileset, loading its texture along the way.
+    pub fn from_tileset<'p>(tileset: Tileset) -> Self {
+        let tileset_image = tileset.image.as_ref().unwrap();
+
+        let texture = {
+            let texture_path = Path::new(&tileset_image.source);
+
+            Texture::from_file(texture_path.to_str().expect("obtaining valid UTF-8 path")).unwrap()
```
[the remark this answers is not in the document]
alexdevteam: Yes, that messy code was the result of a change to make the example work with this codebase.

bjorn:
> The sfml crate is supposed to vendor its own version of sfml, but for some reason still requires those packages as dependencies. I've had issues with that as well in the past, I'll create an issue in the sfml-rs repo.

Hmm, according to their documentation they expect you to just install SFML and a suitable compiler first.

alexdevteam:
> Hmm, according to their documentation they expect you to just install SFML and a suitable compiler first.

See [this PR](https://github.com/jeremyletang/rust-sfml/pull/242). I don't think that *should* apply any more, at least on master. (0.16 does not vendor CSFML).

bjorn:
I just realized that the autobuild is failing because of the SFML dependency, even though actually it is only used by one of the examples, which isn't actually being compiled. Is there any way to make this dependency only install when trying to compile that example?

bjorn:
@alexdevteam Would you mind allowing me to push into your branch (the pull request setting), so that I could just push https://github.com/bjorn/rs-tiled/commit/b547ffa8a74479f1927419fdf083cca24a8affc1?

alexdevteam:
Sure, but I can't find the option :(
I've already looked [here](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/working-with-forks/allowing-changes-to-a-pull-request-branch-created-from-a-fork), but there's nothing in that place. Where is the setting?

bjorn:
> I've already looked [here](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/working-with-forks/allowing-changes-to-a-pull-request-branch-created-from-a-fork), but there's nothing in that place. Where is the setting?

It should be exactly there, so not sure why you can't see it. Maybe you could just cherry-pick this change for now.

alexdevteam:
Okay there we go, for some reason the setting isn't visible if the repo is in an organization. Moved it to my account and now you should be able to push changes just fine.

bjorn:
> Okay there we go, for some reason the setting isn't visible if the repo is in an organization. Moved it to my account and now you should be able to push changes just fine.

Ah, ok! So, that was what the "On user-owned forks" in the docs was about. That wasn't exactly standing out, heh.

examples/sfml/tilesheet.rs
bjorn: In my changes, I've removed the 0 check, but kept the gid for now. We can revisit that when we adjust the example to support multiple tilesets, because in that case we need to find a nice way to call `tile_rect` on the right `Tilesheet` instance, and this also involves using the `first_gid`. And... we might resolve that during loading anyway.

bjorn:
> Is there any way to make this dependency only install when trying to compile that example?

If there is no way to do this, I would be fine with adding the installation of `libsfml-dev` to the autobuild, btw.

alexdevteam:
This is strange. Build is failing as if it the crate modules were private.
Might it be the rust-cache action, which is giving irrelevant data to the CI?

alexdevteam:
image
image

Any idea why this could be happening? rust-analyzer complains as well but cargo builds without any issues on my local system.

alexdevteam:
Alright, fixed, that should do it for the CI. We now have two workflows; One for the library and one for the SFML example. The SFML example one only should run when there has been some change in the examples/sfml directory.

bjorn (approved):
Looks fine to me! I'm not really sure about splitting off the build for the SFML example though, since we'd want to make sure it compiles also when making changes to the library anyway.

G-0-0:
Thanks for this example, it's been a great boost to get going with tiled quickly. I've just got a minor suggestion to improve usability:
```diff
-    let map = Map::parse_file(Path::new(MAP_PATH)).unwrap();
+    let mut path = std::fs::canonicalize(".").expect("current dir does not exist");
+    path.push("assets");
+    while !path.exists() {
+        path.pop();
+        if !path.pop() {
+            println!("ERROR: reached file-system root without finding the assets directory");
+            return
+        }
+        path.push("assets");
+    }
+    path.pop();
+    let map = Map::parse_file(path.join(MAP_PATH)).unwrap();
```

Something like this to allow the example to be run from any directory within the project. Makes it easier to poke around with the example if (like me) you want to use it to understand how rs-tiled works.

bjorn:
> Something like this to allow the example to be run from any directory within the project. Makes it easier to poke around with the example if (like me) you want to use it to understand how rs-tiled works.

Thanks for trying out the example @G-0-0, and I'm glad it helped you! Regarding your patch, I'm not sure this approach sets a good example. First of all I'd like to know, how were you running the example? Normally I'd expect people to execute `cargo run --example sfml`, which should have the correct working directory for the example to run.

I've also taken a look at how Bevy finds assets, and they use the `CARGO_MANIFEST_DIR` environment variable. See [their comment here](https://github.com/bevyengine/bevy/blob/4134577e646e0d1017064780bca34bb3fa096216/examples/asset/asset_loading.rs#L18-L25) and [the implementation here](https://github.com/bevyengine/bevy/blob/f3de12bc5e03c58a9b863943b8db84a321422bec/crates/bevy_asset/src/io/file_asset_io.rs#L38-L51), which falls back to the executable's path if the environment variable isn't available. This looks a bit cleaner to me, so I'm curious if it would also work for you.

alexdevteam:
> Thanks for trying out the example @G-0-0, and I'm glad it helped you! Regarding your patch, I'm not sure this approach sets a good example. First of all I'd like to know, how were you running the example? Normally I'd expect people to execute `cargo run --example sfml`, which should have the correct working directory for the example to run.

Have to agree with @bjorn here. Right now my priority with this PR is to get the lowest amount of boilerplate possible, so adding code like that wouldn't show anything to the end user about the library, and furthermore, I expect this code to be run with cargo. As such, I don't think it's appropriate to use it. However, I could replace the `unwrap` in `parse_file` with an `expect` giving the example runner a bit more feedback about why it couldn't load.

G-0-0:
Thanks both for the responses.

> Normally I'd expect people to execute cargo run --example sfml, which should have the correct working directory for the example to run.

I am running this with cargo:
```
$ cd examples/sfml/
$ cargo r --example sfml
    Finished dev [unoptimized + debuginfo] target(s) in 0.01s
     Running `<redacted>/rs-tiled/target/debug/examples/sfml`
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: Other("Map file not found: \"assets/tiled_base64_external.tmx\"")', examples/sfml/main.rs:92:52
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

Anyway, my intention wasn't to stop this getting merged, so please go ahead and I can address this afterwards. Also I wasn't aware of that method of finding the executable path, so thanks for sharing the Bevy example, it's a lot nicer.

bjorn:
> I am running this with cargo:

Ah, so, in case you were not aware, you don't need to `cd examples/sfml/` to run the example (and personally, I was not aware it worked even if you did!). Anyway, in that case I think we should indeed ensure the example works either way.

alexdevteam:
That should fix it.

alexdevteam:
A few TODOs for this PR:

- [ ] Add a prettier map
- [ ] Fix CI (again)

alexdevteam:
Blocked by #135.

examples/main.rs
```diff
@@ -1,7 +1,50 @@
-use tiled::Map;
+use std::path::PathBuf;
+
+use tiled::{FilesystemResourceCache, Map};
 
 fn main() {
-    let map = Map::parse_file("assets/tiled_base64_zlib.tmx").unwrap();
-    println!("{:?}", map);
-    println!("{:?}", map.tileset_by_gid(22));
```
alexdevteam, line 6: I messed up, right? Why are things like these counted as changes in respect to master?

bjorn (approved):
This is looking pretty good now I think. Is there a reason the PR is still marked as Draft?

bjorn (approved):

alexdevteam:
> This is looking pretty good now I think. Is there a reason the PR is still marked as Draft?

I'd like a prettier map to show instead of the one we have now.

alexdevteam:
Alright, merging then.

PR #135, opened 2022-01-26
Hide GIDs as internal details

### LINKED ISSUE #123: Erase GIDs from the public interface
Global tile IDs appear frequently in the codebase but I feel like they are not explained throughout, and they lack their own type so they can be confused with local (tileset) IDs.

alexdevteam:
Closes #123. WIP.

### DIFF
--- a/src/map.rs
+++ b/src/map.rs
@@ -7,6 +7,7 @@ use crate::{
     layers::{ImageLayer, Layer},
     objects::ObjectGroup,
     properties::{parse_properties, Color, Properties},
+    tile::Tile,
     tileset::Tileset,
     util::{get_attrs, parse_tag},
 };
@@ -74,6 +75,28 @@ impl Map {
         let file = File::open(path.as_ref())
             .map_err(|_| TiledError::Other(format!("Map file not found: {:?}", path.as_ref())))?;
         Self::parse_reader(file, Some(path.as_ref()))
+    }
+}
+
+impl Map {
+    pub(crate) fn tile_by_gid(&self, gid: u32) -> Option<Tile> {
+        let mut max_gid_ts = None;
+        for tileset in self.tilesets.iter() {
+            if max_gid_ts
+                .map(|(gid, _)| tileset.first_gid > gid)
+                .unwrap_or(false)
+                && tileset.first_gid <= gid
+            {
+                max_gid_ts = Some((tileset.first_gid, tileset));
+            }
+        }
+
+        if let Some((gid, ts)) = max_gid_ts {
+            let local_id = gid - ts.first_gid;
+            ts.get_tile(local_id)
+        } else {
+            None
+        }
     }
 
     fn parse_xml<R: Read>(
@@ -146,19 +169,6 @@ impl Map {
             background_color: c,
             infinite: infinite.unwrap_or(false),
         })
-    }
-
-    /// This function will return the correct Tileset given a GID.
-    pub fn tileset_by_gid(&self, gid: u32) -> Option<&Tileset> {
-        let mut maximum_gid: i32 = -1;
-        let mut maximum_ts = None;
-        for tileset in self.tilesets.iter() {
-            if tileset.first_gid as i32 > maximum_gid && tileset.first_gid <= gid {
-                maximum_gid = tileset.first_gid as i32;
-                maximum_ts = Some(tileset);
-            }
-        }
-        maximum_ts
     }
 }
 
--- a/src/tileset.rs
+++ b/src/tileset.rs
@@ -17,7 +17,7 @@ use crate::util::*;
 #[derive(Debug, PartialEq, Clone)]
 pub struct Tileset {
     /// The GID of the first tile stored.
-    pub first_gid: u32,
+    pub(crate) first_gid: u32,
     pub name: String,
     pub tile_width: u32,
     pub tile_height: u32,
@@ -34,7 +34,7 @@ pub struct Tileset {
     /// - Source: [tiled issue #2117](https://github.com/mapeditor/tiled/issues/2117)
     /// - Source: [`columns` documentation](https://doc.mapeditor.org/en/stable/reference/tmx-map-format/#tileset)
     pub image: Option<Image>,
-    pub tiles: Vec<Tile>,
+    special_tiles: HashMap<u32, Tile>,
     pub properties: Properties,
 
     /// Where this tileset was loaded from.
@@ -79,6 +79,20 @@ impl Tileset {
         Tileset::new_external(reader, first_gid, Some(path.as_ref()))
     }
 
+    /// Gets a clone of the tile with the local ID specified, if it exists within this tileset.
+    pub fn get_tile<'s: 't, 't>(&'s self, id: u32) -> Option<Tile> {
+        if let Some(tile) = self.special_tiles.get(&id) {
+            Some(tile.clone())
+        } else {
+            Some(Tile {
+                id,
+                ..Default::default()
+            })
+        }
+    }
+}
+
+impl Tileset {
     pub(crate) fn parse_xml<R: Read>(
         parser: &mut EventReader<R>,
         attrs: Vec<OwnedAttribute>,
@@ -140,7 +154,7 @@ impl Tileset {
                 ("columns", columns, |v:String| v.parse().ok()),
             ],
            required: [
-                ("firstgid", first_gid, |v:String| v.parse().ok()),
+            ("firstgid", first_gid, |v:String| v.parse().ok()),
                 ("name", name, |v| Some(v)),
                 ("tilewidth", width, |v:String| v.parse().ok()),
                 ("tileheight", height, |v:String| v.parse().ok()),
@@ -239,7 +253,7 @@ impl Tileset {
         prop: TilesetProperties,
     ) -> Result<Self, TiledError> {
         let mut image = Option::None;
-        let mut tiles = Vec::new();
+        let mut tiles = HashMap::new();
         let mut properties = HashMap::new();
         parse_tag!(parser, "tileset", {
             "image" => |attrs| {
@@ -251,7 +265,8 @@ impl Tileset {
                 Ok(())
             },
             "tile" => |attrs| {
-                tiles.push(Tile::new(parser, attrs, prop.path_relative_to.as_ref().and_then(|p| Some(p.as_path())))?);
+                let tile = Tile::new(parser, attrs, prop.path_relative_to.as_ref().and_then(|p| Some(p.as_path())))?;
+                tiles.insert(tile.id, tile);
                 Ok(())
             },
         });
@@ -273,7 +288,7 @@ impl Tileset {
             columns,
             tilecount: prop.tilecount,
             image,
-            tiles,
+            special_tiles: tiles,
             properties,
             source: prop.source,
         })
--- a/src/layers.rs
+++ b/src/layers.rs
@@ -5,15 +5,16 @@ use xml::{attribute::OwnedAttribute, EventReader};
 use crate::{
     error::TiledError,
     image::Image,
+    map::Map,
     properties::{parse_properties, Properties},
+    tile::Tile,
     util::*,
 };
 
-/// Stores the proper tile gid, along with how it is flipped.
-// Maybe PartialEq and Eq should be custom, so that it ignores tile-flipping?
-#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+/// Contains all the data about a tile in a layer.
+#[derive(Debug, Clone, PartialEq)]
 pub struct LayerTile {
-    pub gid: u32,
+    pub tile: Tile,
     pub flip_h: bool,
     pub flip_v: bool,
     pub flip_d: bool,
@@ -25,20 +26,36 @@ const FLIPPED_DIAGONALLY_FLAG: u32 = 0x20000000;
 const ALL_FLIP_FLAGS: u32 =
     FLIPPED_HORIZONTALLY_FLAG | FLIPPED_VERTICALLY_FLAG | FLIPPED_DIAGONALLY_FLAG;
 
-impl LayerTile {
-    pub fn new(id: u32) -> LayerTile {
-        let flags = id & ALL_FLIP_FLAGS;
-        let gid = id & !ALL_FLIP_FLAGS;
+/// Stores the internal tile gid about a layer tile, along with how it is flipped.
+struct LayerTileInner {
+    gid: u32,
+    flip_h: bool,
+    flip_v: bool,
+    flip_d: bool,
+}
+
+impl LayerTileInner {
+    fn new(bits: u32) -> Self {
+        let flags = bits & ALL_FLIP_FLAGS;
+        let gid = bits & !ALL_FLIP_FLAGS;
         let flip_d = flags & FLIPPED_DIAGONALLY_FLAG == FLIPPED_DIAGONALLY_FLAG; // Swap x and y axis (anti-diagonally) [flips over y = -x line]
         let flip_h = flags & FLIPPED_HORIZONTALLY_FLAG == FLIPPED_HORIZONTALLY_FLAG; // Flip tile over y axis
         let flip_v = flags & FLIPPED_VERTICALLY_FLAG == FLIPPED_VERTICALLY_FLAG; // Flip tile over x axis
 
-        LayerTile {
+        Self {
             gid,
             flip_h,
             flip_v,
             flip_d,
         }
+    }
+
+    /// Gets a clone of the [`Tile`] associated to this [`LayerTile`], given the map that the tile is in.
+    ///
+    /// If the layer tile has an invalid ID (which can happen if, for instance, the tile isn't from the map given),
+    /// this will return [`None`].
+    pub fn get_associated_tile(&self, map: &Map) -> Option<Tile> {
+        map.tile_by_gid(self.gid)
     }
 }
 
@@ -53,7 +70,7 @@ pub struct Layer {
     pub offset_y: f32,
     /// The tiles are arranged in rows. Each tile is a number which can be used
     ///  to find which tileset it belongs to and can then be rendered.
-    pub tiles: LayerData,
+    tiles: LayerData,
     pub properties: Properties,
     pub layer_index: u32,
     /// The ID of the layer, as shown in the editor.
--- a/src/tile.rs
+++ b/src/tile.rs
@@ -11,7 +11,7 @@ use crate::{
     util::{get_attrs, parse_animation, parse_tag},
 };
 
-#[derive(Debug, PartialEq, Clone)]
+#[derive(Debug, PartialEq, Clone, Default)]
 pub struct Tile {
     pub id: u32,
     pub image: Option<Image>,


bjorn:
When doing this change, it is important that we use the "flyweight" pattern, because the `Tile` is way too heavy a data structure to be copying around, both for storing layer data (so it can't be a member of `LayerTile`) as well as when looking them up from the tileset (so we can't just return `Option<Tile>`).

We need the `LayerTile` to contain a reference to a `Tile` that can be quickly looked up. The `gid` is such a reference, but it is not really quick to look up due to the need to first find which tileset is being referenced. I guess in Rust any kind of reference is going to introduce the problem of lifetimes. We need to ensure that the referenced tileset stays alive at least as long as any `LayerTile` values.

Tiled itself has just used `Tile*` for quite some years, but this meant tiles could not be removed from tilesets without also making sure to clear any such pointers, which was of course a tricky business since these pointers existed not only in tile layers and tile objects, but also in the current tile stamp, preview layers, etc. I've tried to improve this situation by largely switching to `(Tileset*, int)` for referencing tiles, because it is much easier to keep the tileset alive (using shared pointers in various places). Then of course when looking up tiles from the tileset, generally we need to handle the case that the tile might not longer be there.

Talking about shared pointers, one option might be to use `(Arc<Tileset>, u32)`, but I think the overhead of the reference counting at that level would be way too much.

Or we could just use numbers, for example `TileHandle { tileset_index: u32, tile_id: u32 }`, where the `tileset_index` would be just an index into the array of tilesets. Based on this, the `Map` could provide:
```rust
fn get_tile(&self, &handle: TileHandle) -> Option<&Tile> {
```
I think the main downside is that we then always need the `Map` to interpret a tile layer, but I think that makes sense if the map owns the tilesets, and at least currently it also specifies a lot of other relevant parameters.

Any other thoughts on this, especially from more experienced Rust developers?

alexdevteam:
The issue with that is that you don't own every Tile, only the ones with special properties, so you could be returning either A) a reference to an owned tile or B) a reference to a local variable in get_tile.
I tried to achieve that exact interface but wasn't able to; Maybe there's some way, I haven't spent much time on it.

bjorn:
> The issue with that is that you don't own every Tile, only the ones with special properties, so you could be returning either A) a reference to an owned tile or B) a reference to a local variable in get_tile.

That is mainly a loader issue. I think we should actually create a `Tile` instance for every tile, also those that do not have additional properties in the stored file. As an added bonus, we can store the texture coordinates for each tile there, so that they don't need to be calculated during rendering.

alexdevteam:
Should be easy then. I'll do it that way then.

alexdevteam:
Alright, interface is finally taking shape. Some quirks which need attention:

- [ ] Infinite maps require a reimplementation.
- [ ] The current interface allows obtaining references to tiles via `Map::get_tile`, which takes in a layer index and x & y coordinates. Figure out some way to remove that layer index.
- [ ] Some tests fail.

Test status:
```sh
running 14 tests
test test_image_layers ... ok
test test_ldk_export ... ok
test test_object_group_property ... ok
test test_object_property ... ok
test test_parallax_layers ... ok
test test_flipped ... FAILED
test test_infinite_tileset ... FAILED
test test_sources ... ok
test test_layer_property ... ok
test test_tileset_property ... ok
test test_just_tileset ... ok
test test_tile_property ... FAILED
test test_external_tileset ... ok
test test_gzip_and_zlib_encoded_and_raw_are_the_same ... FAILED
```

alexdevteam:
The only reasonable way I have thought of to remove the layer index in `Map::get_tile` is to replace the `layers` public member with a property that gives a layer facade for each index, which includes the `Layer` as well as a reference to the parent map. This would allow to put the `get_tile` method there without any issues.

The reason a simple `get_tile(map: &Map) -> Tile` in LayerTile won't make it is because that introduces invalid states by allowing users to give a map which isn't the one the tile came from as parameter.

bjorn:
What I was trying to say with the `TileHandle` struct, is that what we really want to do is to transform the global tile IDs to a structure where we no longer have to look up the tileset each tile we want to access the actual tile. This is for two reasons:

* The global tile IDs are just a mechanism of the file format. The map could be stored any other way, and in memory it does not make much sense to keep the global tile IDs.
* The translation from global tile IDs to tile references has an overhead that we do not want to repeat every time we look up a tile, for example to render that tile. I'm talking especially about the `Map.get_tile_by_gid` function here, since I guess the hash map lookup in `Tileset.get_tile_by_gid` should be fine.

I'm wondering, why did you remove `id` from `Tile` and added it to `LayerTileRef`? If we just keep storing the `id` in each `Tile` instance, we don't need it in `LayerTileRef`. Or, if we keep the `id` also in `LayerTileRef`, then `LayerTileRef` does not need the `&'map Tile`, because it can be obtained from the `&'map Tileset`.

alexdevteam:
> I'm wondering, why did you remove `id` from `Tile` and added it to `LayerTileRef`? If we just keep storing the `id` in each `Tile` instance, we don't need it in `LayerTileRef`.

To avoid having duplicate data in the tileset hashmap, since the keys are already the tile IDs.

> Or, if we keep the `id` also in `LayerTileRef`, then `LayerTileRef` does not need the `&'map Tile`, because it can be obtained from the `&'map Tileset`.

Right, but map already needs to look up the tile anyways, so it can save a second lookup in the tileset.

bjorn:
> To avoid having duplicate data in the tileset hashmap, since the keys are already the tile IDs.

Often it makes sense to store this information since the reverse lookup in the map is expensive, but maybe we can avoid the need for doing reverse lookups (so, it's fine if we indeed never need to get the `id` while just having a `&Tile`).

> Right, but map already needs to look up the tile anyways, so it can save a second lookup in the tileset.

It didn't really have to do that. I've pushed a change that removes the `tile` member from `LayerTileRef` (and delays the lookup in the hash map until the `&Tile` is actually requested).

A few issues remain that should be addressed:

* We still need to do a tileset lookup every time a tile is queried.
* We can't have `Map.get_tile` taking a layer index. If we could replace the `LayerTileGid` with the `LayerTileRef`... but I guess there is no way to do that in Rust due to lifetimes?

src/map.rs
```diff
@@ -89,11 +88,11 @@ impl Map {
...
 
 impl Map {
-    pub(crate) fn get_tile_by_gid(&self, gid: Gid) -> Option<(&Tile, &Tileset)> {
+    pub(crate) fn get_tileset_for_gid(&self, gid: Gid) -> Option<&Tileset> {
         self.tilesets
             .iter()
-            .filter_map(|ts| ts.get_tile_by_gid(gid).map(|tile| (tile, ts)))
-            .nth(0)
+            .rev()
+            .find(|ts| ts.first_gid <= gid.0)
```
alexdevteam, line 95: This assumes that tilesets are ordered by first GID. Should be documented.

alexdevteam:
> It didn't really have to do that. I've pushed a change that removes the tile member from LayerTileRef (and delays the lookup in the hash map until the &Tile is actually requested).

The issue here is that `get_tileset_for_gid` is now more of a `get_possible_tileset_for_gid` since there's no guarantee that the tile will actually be in that tileset (think image collections or invalid GIDs)

src/map.rs
bjorn: This is documented at https://doc.mapeditor.org/en/stable/reference/tmx-map-format/#tmx-tileset:
> If there are multiple <tileset> elements, they are in ascending order of their firstgid attribute.

bjorn:
> The issue here is that `get_tileset_for_gid` is now more of a `get_possible_tileset_for_gid` since there's no guarantee that the tile will actually be in that tileset (think image collections or invalid GIDs)

Right, but that is irrelevant. It's still the only tileset that is supposed to contain the tile for that `gid`. It makes no sense to look for it in any other tileset.

alexdevteam:
Alright, all the data is in its place. The issue now is the interface. We need to somehow expose layers and allow the user to fetch tiles as well as tilesets, which is much easier said than done. I really can't see any other way to achieve it without putting wrappers on literally all layer types.

alexdevteam:
Done, and it ended a lot cleaner that I thought! I defined a generic type `LayerWrapper<'map, DataT>` That holds a `&'map Map` as well as a `&'map DataT`, and then instantiated it for all the layer types. There isn't much repeated code, if any.

src/map.rs
```diff
@@ -96,30 +185,31 @@ impl Map {
...
-        let mut tilesets = Vec::new();
         let mut layers = Vec::new();
         let mut properties = HashMap::new();
+        let mut tilesets = Vec::new();
         parse_tag!(parser, "map", {
             "tileset" => |attrs| {
-                tilesets.push(Tileset::parse_xml(parser, attrs, source_path)?);
+                let res = Tileset::parse_xml_in_map(parser, attrs, map_path)?;
+                tileset_cache.get_or_insert(&res.tileset_path, res.tileset);
+                tilesets.push(TilesetRef{first_gid: res.first_gid, path: res.tileset_path});
```
bjorn, line 196: I think there are two problems here:

* The tileset, be it embedded or external, is still always parsed, even if it is already part of the cache.
* We insert embedded tilesets in the cache using an empty path.

I'm thinking `EmbeddedParseResult` should have an `Option<Tileset>` and an `Option<PathBuf>`, or we should use an enum to say it's either an embedded tileset, or a reference to an external one.

In case it's a reference, we should not immediately call `Tileset::new_external` from `parse_xml_reference`, but rather just return the `tileset_path`, such that we can obtain the tileset from the cache (or load it, if not present) in `Map::parse_xml`. To enable this, `TilesetCache::get_or_insert` could probably be changed to `get_or_insert_with` and take function.

bjorn:
I wonder if it makes sense to already name the "TilesetCache" a little more generically, since we'll have the same issue when adding support for object templates (https://github.com/mapeditor/rs-tiled/pull/96#discussion_r790580315). Maybe it should be a `ResourceCache`.

Also wondering, is this trait going to work when resources are loaded through async IO, like HTTP?

alexdevteam:
No, it won't work since it'd require to be async (and we'd need an async trait, which would require an extra dependency). We can worry about that later on though.

src/map.rs
alexdevteam: > To enable this, TilesetCache::get_or_insert could probably be changed to get_or_insert_with and take a function.

It's a bit more complex than that since `new_external` returns a Result, and we can't handle the result inside of the function we pass in. Hmm

alexdevteam: Fixed in latest commit

src/map.rs
```diff
@@ -192,8 +185,20 @@ impl Map {
...
+                let path = match res.result_type {
+                    EmbeddedParseResultType::ExternalReference { tileset_path } => {
+                        let path = crate::ResourcePath::External{ path: tileset_path.clone() };
+                        let file = File::open(&tileset_path).map_err(|err| TiledError::CouldNotOpenFile{path: tileset_path.clone(), err })?;
+                        tileset_cache.get_or_try_insert_tileset_with(path.clone(), || Tileset::new_external(file, &tileset_path))?;
+                        path
+                    }
+                    EmbeddedParseResultType::Embedded { tileset } => {
+                        let path = ResourcePath::Embedded { container_path: map_path.to_owned(), index: tilesets.len()};
+                        tileset_cache.get_or_insert_tileset(path.clone(), tileset);
```
bjorn, line 197: I don't really understand this. Why did you introduce `ResourcePath::Embedded`? What's the point of caching embedded tilesets, given that no other map could reference them?

alexdevteam:
Decided to leave fixing the infinite layers interface for another PR.

alexdevteam:
> The `gid` attribute from objects is now hidden away and there is no way to access the tile it is referencing.

Forgot about this, will fix now

src/cache.rs
```diff
@@ -0,0 +1,64 @@
...
+        f: F,
+    ) -> Result<&Tileset, E>
+    where
+        F: FnOnce() -> Result<Tileset, E>,
+    {
+        if !self.tilesets.contains_key(&path) {
+            Ok(self.tilesets.entry(path).or_insert(f()?))
+        } else {
+            Ok(self.tilesets.get(&path).unwrap())
+        }
```
alexdevteam, line 62: Is there a better way to do this?

bjorn: I guess since loading the tileset might fail we can't use `or_insert_with`. I've tried:
```rust
        Ok(self.tilesets.entry(path).or_insert_with(|| {
            f()?
        }))
```
But it results in something I don't understand:
```
error[E0277]: the trait bound `tileset::Tileset: FromResidual<Result<Infallible, E>>` is not satisfied
   --> src/cache.rs:63:53
    |
63  |           Ok(self.tilesets.entry(path).or_insert_with(|| {
    |  ______________________________________--------------_^
    | |                                      |
    | |                                      required by a bound introduced by this call
64  | |             f()?
65  | |         }))
    | |_________^ the trait `FromResidual<Result<Infallible, E>>` is not implemented for `tileset::Tileset`
    |
note: required by `from_residual`
   --> /home/bjorn/.rustup/toolchains/stable-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/library/core/src/ops/try_trait.rs:339:5
    |
339 |     fn from_residual(residual: R) -> Self;
    |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```
Reading [`FromResidual`](https://doc.rust-lang.org/std/ops/trait.FromResidual.html) hasn't really helped me, but I doubt it's going to achieve what we want, namely that a potential error in the callback is returned by `or_insert_with`.

But, we could at least avoid the double hash map lookup:

```suggestion
        Ok(match self.tilesets.entry(path) {
            Entry::Occupied(entry) => entry.into_mut(),
            Entry::Vacant(entry) => entry.insert(f()?),
        })
```

alexdevteam: > But, we could at least avoid the double hash map lookup

Ah, that's it. Didn't know Entry was an enum.

bjorn: > You can't use `?` because the parent closure should return `T`, not `Result<T, E>`.

Right, and apparently for some types one could implement `FromResidual<Result<Infallible, E>>` so that you get a valid object even in the case of error.

> Ah, that's it. Didn't know Entry was an enum.

Right, in fact the `Entry` seems to be very much like your `LayerTile`, in that it enables getting references to elements with the lifetime of their container.

src/tileset.rs
```diff
@@ -126,145 +102,170 @@ impl Tileset {
...
-                ("tilecount", tilecount, |v:String| v.parse().ok()),
                 ("columns", columns, |v:String| v.parse().ok()),
             ],
             required: [
+                ("tilecount", tilecount, |v:String| v.parse().ok()),
                 ("name", name, |v| Some(v)),
                 ("tilewidth", width, |v:String| v.parse().ok()),
                 ("tileheight", height, |v:String| v.parse().ok()),
             ],
             TiledError::MalformedAttributes("tileset must have a firstgid, name tile width and height with correct types".to_string())
```
bjorn, line 205: Should probably mention `tilecount` now, and should not mention `firstgid`.

And actually I don't see why `name` should be required, it can just default to an empty string.

bjorn (changes requested):
Nice work overall! I left some comments based on a full review. :-)

src/tileset.rs
alexdevteam: The `name` requirement is a separate issue; I'll fix the message though.

src/objects.rs
```diff
@@ -18,9 +19,9 @@ pub enum ObjectShape {
 }
 
 #[derive(Debug, PartialEq, Clone)]
-pub struct Object {
+pub struct ObjectData {
     pub id: u32,
-    pub gid: u32,
+    tileset_idx_tile_id: Option<(u32, u32)>,
```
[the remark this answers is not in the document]
alexdevteam: Oh, perfect then, wasn't using it because I didn't know that flipping was possible here.

src/map.rs
```diff
@@ -96,33 +186,94 @@ impl Map {
         );
 
         let infinite = infinite.unwrap_or(false);
-        let source_path = map_path.and_then(|p| p.parent());
 
-        let mut tilesets = Vec::new();
+        // Since we can only parse sequentally, we cannot ensure tilesets are parsed before tile
+        // layers. The issue here is that tile layers require tileset data in order to set
+        // [`LayerTileData`] correctly. As such, we parse layer processing data into a vector as well as tilesets
+        // and then, when all the tilesets are parsed, we actually construct each layer.
```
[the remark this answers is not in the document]
bjorn: You're right, apparently this is not documented. I shall need to add a note about that...

bjorn:
> Why would they count towards ownership?

They don't count towards _ownership_, but they're still counted, because as long as there is a weak reference present, the actual memory allocated for the reference-counted object cannot be freed (it would leave the remaining weak pointers without any means to know whether the memory they're pointing at is still valid or not).

alexdevteam:
Well then, we'd be able to do `&'map Rc<Tileset>` but not `&'map Tileset`.

src/objects.rs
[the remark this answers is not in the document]
alexdevteam: Where's this in the docs, BTW? I can only find the `gid` attribute but not anything related to flipping in https://doc.mapeditor.org/en/stable/reference/tmx-map-format/#tmx-object.

bjorn:
> Well then, we'd be able to do `&'map Rc<Tileset>` but not `&'map Tileset`.

I don't understand, that doesn't help at all, would it? I've pushed a change that passes the tests for me. Am I missing something?

src/objects.rs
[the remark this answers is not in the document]
bjorn: I guess it's kind of implicit, and it would be good to explicitly say that because this is a GID, it might also include its [flipping flags](https://doc.mapeditor.org/en/stable/reference/global-tile-ids/#tile-flipping).

bjorn:
> We can't keep direct references to the cache in maps, otherwise we won't be able to modify it while a map is in scope. Same thing goes for `LayerTile`.

I guess we need to include a test where a tile is modified, to make sure we don't make that impossible. :-)

alexdevteam:
> I guess we need to include a test where a tile is modified, to make sure we don't make that impossible. :-)

You mean the cache is modified, right? :p

src/layers/tile/mod.rs
```diff
@@ -0,0 +1,145 @@
...
+    const FLIPPED_VERTICALLY_FLAG: u32 = 0x40000000;
+    const FLIPPED_DIAGONALLY_FLAG: u32 = 0x20000000;
+    const ALL_FLIP_FLAGS: u32 = Self::FLIPPED_HORIZONTALLY_FLAG
+        | Self::FLIPPED_VERTICALLY_FLAG
+        | Self::FLIPPED_DIAGONALLY_FLAG;
+
+    /// Creates a new, **unfinished** [`LayerTileData`]. Unfinished layer tiles have the following
+    /// properties:
+    /// - `tileset_index` is set to an unspecified value.
+    /// - `id` is actually the tile's GID.
```
alexdevteam, line 41: This is wrong now.

examples/main.rs
```diff
@@ -1,7 +1,50 @@
...
+            tiled::LayerType::TileLayer(layer) => match layer.data() {
+                tiled::TileLayerData::Finite(data) => println!(
+                    "Finite tile layer with width = {} and height = {}; ID of tile @ (0,0): {}",
+                    data.width(),
+                    data.height(),
+                    layer.get_tile(0, 0).unwrap().id
+                ),
+                tiled::TileLayerData::Infinite(data) => {
+                    // This is prone to change! Infinite layers will be refactored before 0.10.0
+                    // releases.
```
bjorn, line 29: Hmm, forward looking statements. Do we need to add a disclaimer?

bjorn (approved):
Great work @alexdevteam!

PR #139, opened 2022-02-03
Some Cargo.toml updates

bjorn:
* Added "game-development" to "categories"
* Added "gamedev" to "keywords"
* Updated repository URL
* Updated dependencies
* Removed broken "documentation" URL

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -2,15 +2,14 @@
 
 name = "tiled"
 version = "0.9.5"
-description = "A rust crate for loading in maps created by the Tiled editor"
-repository = "https://github.com/mattyhall/rs-tiled.git"
-# documentation = "http://rust-ci.org/mattyhall/rs-tiled/doc/tiled/"
+description = "A rust crate for loading maps created by the Tiled editor"
+categories = ["game-development"]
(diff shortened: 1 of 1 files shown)


Cargo.toml
```diff
@@ -24,7 +23,7 @@ name = "example"
 path = "examples/main.rs"
 
 [dependencies]
-base64  = "0.10"
-xml-rs  = "0.8"
-libflate = "0.1.18"
-zstd = { version = "0.9", optional = true }
+base64  = "0.13.0"
+xml-rs  = "0.8.4"
```
alexdevteam, line 27: Fixing the minor version is not a good idea. These won't be breaking and we'll get them for free if we just do `0.8`.

alexdevteam (changes requested):

Cargo.toml
bjorn: The only difference between putting `0.8` or `0.8.4` is in the minimum version, in both cases the upper bound is `<0.9.0`. These version bumps are just done using `cargo upgrade` (after installing `cargo-edit`).

alexdevteam: I see.

alexdevteam (approved):

PR #150, opened 2022-02-09
Add layer tint color; Fix `Color` parsing issues

### LINKED ISSUE #85: Parse layers' tint color
It would be great if we could get [the tint color of layer](https://thorbjorn.itch.io/tiled/devlog/120663/tinting-alignment-and-object-references). This would avoid duplicating it in properties (which can cause a bugs). I can try to implement this.

### LINKED ISSUE #148: Color parsing does not take `alpha` into account
The `Color` structure does not keep track of the alpha value of the color, and the `FromStr` implementation of `Color` is wrong since it assumes that colors are in the `#RRGGBB` format (as opposed to `#AARRGGBB`)

### LINKED ISSUE #149: Add `Color` tests
Add tests for parsing colors.

alexdevteam:
Closes #149, closes #85, fixes #148.

### DIFF
--- a/src/properties.rs
+++ b/src/properties.rs
@@ -1,38 +1,61 @@
 use std::{collections::HashMap, io::Read, str::FromStr};
 
-use xml::{EventReader, attribute::OwnedAttribute, reader::XmlEvent};
+use xml::{attribute::OwnedAttribute, reader::XmlEvent, EventReader};
 
 use crate::{
-    error::{ParseTileError, TiledError},
+    error::TiledError,
     util::{get_attrs, parse_tag},
 };
 
 #[derive(Debug, PartialEq, Eq, Copy, Clone)]
 pub struct Color {
+    pub alpha: u8,
     pub red: u8,
     pub green: u8,
     pub blue: u8,
 }
 
 impl FromStr for Color {
-    type Err = ParseTileError;
+    type Err = ();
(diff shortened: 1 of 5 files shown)


alexdevteam:
I thought this would be my first nonbreaking PR, but sadly the addition of the `alpha` member in `Color` makes it breaking. Sad

src/properties.rs
```diff
@@ -1,38 +1,61 @@
...
-            return Ok(Color {
-                red: r.unwrap(),
-                green: g.unwrap(),
-                blue: b.unwrap(),
-            });
+    fn from_str(s: &str) -> Result<Color, Self::Err> {
+        let s = if s.starts_with("#") {
+            &s[1..]
+        } else {
+            return Err(());
```
bjorn, line 25: Actually, taking `s` as a fallback is what we need to do, because for backwards compatibility reasons, the `trans` attribute on the `image` element leaves out the `#` but is otherwise a valid color value.

(and this is actually documented for once ;-))

bjorn (changes requested):
Looking good, just one small issue.

bjorn (dismissed):

bjorn:
Wait a sec, you're still missing the changelog entry!

bjorn (approved):

PR #152, opened 2022-02-10
Better errors

### LINKED ISSUE #124: Make TiledError non-exhaustive
That way backwards-compatibility will be preserved when adding new error types.

### LINKED ISSUE #142: Better errors
Remove `TiledError::Other`, remove the arbitrary string in `MalformedAttributes` and `PrematureEnd`, and replace them with processable errors.

alexdevteam:
Closes #142 and closes #124.
Replaces all occurrences of `TiledError::Other` for new error variants and removes `ParseTileError`, among other changes.

### DIFF
--- a/src/error.rs
+++ b/src/error.rs
@@ -8,6 +8,7 @@ pub enum ParseTileError {
 
 /// Errors which occured when parsing the file
 #[derive(Debug)]
+#[non_exhaustive]
 pub enum TiledError {
     /// A attribute was missing, had the wrong type of wasn't formated
     /// correctly.
--- a/src/error.rs
+++ b/src/error.rs
@@ -34,6 +34,9 @@ pub enum TiledError {
     /// 
     /// [`PropertyValue`]: crate::PropertyValue
     InvalidPropertyValue,
+    /// Found an unknown property value type while parsing a [`PropertyValue`].
+    /// 
+    /// [`PropertyValue`]: crate::PropertyValue
     UnknownPropertyType{name: String},
 }
 
@@ -79,7 +82,6 @@ impl fmt::Display for TiledError {
     }
 }
 
-// This is a skeleton implementation, which should probably be extended in the future.
 impl std::error::Error for TiledError {
     fn source(&self) -> Option<&(dyn std::error::Error + 'static)> {
         match self {
--- a/src/error.rs
+++ b/src/error.rs
@@ -13,9 +13,8 @@ pub enum TiledError {
     /// A attribute was missing, had the wrong type of wasn't formated
     /// correctly.
     MalformedAttributes(String),
-    /// An error occured when decompressing using the
-    /// [flate2](https://github.com/alexcrichton/flate2-rs) crate.
-    DecompressingError(std::io::Error),
+    /// An error occured when using I/O functionality (e.g. decompression or file reading)
+    IoError(std::io::Error),
     Base64DecodingError(base64::DecodeError),
     XmlDecodingError(xml::reader::Error),
     PrematureEnd(String),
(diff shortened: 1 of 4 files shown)


src/error.rs
```diff
@@ -62,12 +69,20 @@ impl fmt::Display for TiledError {
...
+            TiledError::InvalidEncodingFormat { encoding, compression } => 
+                write!(
+                    fmt,
+                    "Unknown encoding or compression format or invalid combination of both (for tile layers): {} encoding with {} compression",
+                    encoding.as_deref().unwrap_or("no"),
+                    compression.as_deref().unwrap_or("no")
+                ),
+            TiledError::InvalidPropertyValue => write!(fmt, "Found invalid property value"),
+            TiledError::UnknownPropertyType { name } =>
+                write!(fmt, "Found unknown property value type '{}'", name),
```
bjorn, line 81: In this and the above case, I would leave out the "Found" from the error message, which seems unnecessary and inconsistent.

src/layers/tile/util.rs
```diff
@@ -10,18 +10,24 @@ pub(crate) fn parse_data_line(
     parser: &mut impl Iterator<Item = XmlEventResult>,
     tilesets: &[MapTilesetGid],
 ) -> Result<Vec<Option<LayerTileData>>, TiledError> {
-    match (encoding, compression) {
+    match (encoding.as_deref(), compression.as_deref()) {
         (None, None) => {
-            return Err(TiledError::Other(
-                "XML format is currently not supported".to_string(),
-            ))
+            return Err(TiledError::InvalidEncodingFormat {
```
bjorn, line 15: Here, the previous error was more useful. If there is no encoding specified, the original "XML" encoding should be assumed, where each tile gets its own `<tile />` element. Since this format is marked deprecated in Tiled, I would personally remove the "currently", since that indicates there would be plans to support it in the future.

I would suggest to tweak `fmt::Display for TiledError` to be a little more specific again in case neither is provided.

And wen can actually remove this special `(None, None)` case here, cause it is the same as the `_` below.

src/layers/tile/util.rs
```diff
@@ -10,18 +10,24 @@ pub(crate) fn parse_data_line(
...
+            return parse_base64(parser).map(|v| convert_to_tiles(&v, tilesets))
+        }
+        (Some("csv"), None) => return decode_csv(parser, tilesets),
+        (Some(_), None) => {
+            return Err(TiledError::InvalidEncodingFormat {
+                encoding,
+                compression,
+            })
+        }
+        (Some(e), Some(c)) => match (e, c) {
```
bjorn, line 30: I would suggest to use `(Some("base64"), c)` here, cause then we only need to match on `c` afterwards and can merge the `(Some("base64"), None)` case in here, and have only a single place where we call to `parse_base64`.

src/properties.rs
```diff
@@ -74,32 +74,27 @@ impl PropertyValue {
         match property_type.as_str() {
             "bool" => match value.parse() {
                 Ok(val) => Ok(PropertyValue::BoolValue(val)),
-                Err(err) => Err(TiledError::Other(err.to_string())),
+                Err(_) => Err(TiledError::InvalidPropertyValue),
```
bjorn, line 77: It is not helpful to include the original error or its string representation?

bjorn (commented):
Looks pretty good, though I'd suggest some changes.

src/layers/tile/util.rs
alexdevteam: I know, but since this was a PR for errors I decided not to fix it until later

PR #153, opened 2022-02-11
Simplify internals

alexdevteam:
Simplify some expressions and move some functions around. Non-breaking.

### DIFF
--- a/src/layers/tile/util.rs
+++ b/src/layers/tile/util.rs
@@ -15,17 +15,14 @@ pub(crate) fn parse_data_line(
 
         (Some("base64"), None) => parse_base64(parser).map(|v| convert_to_tiles(&v, tilesets)),
         (Some("base64"), Some("zlib")) => parse_base64(parser)
-            .map(|data| std::io::Cursor::new(data))
-            .and_then(|reader| process_decoder(libflate::zlib::Decoder::new(reader)))
+            .and_then(|data| process_decoder(libflate::zlib::Decoder::new(&data[..])))
             .map(|v| convert_to_tiles(&v, tilesets)),
         (Some("base64"), Some("gzip")) => parse_base64(parser)
-            .map(|data| std::io::Cursor::new(data))
-            .and_then(|reader| process_decoder(libflate::gzip::Decoder::new(reader)))
+            .and_then(|data| process_decoder(libflate::gzip::Decoder::new(&data[..])))
             .map(|v| convert_to_tiles(&v, tilesets)),
         #[cfg(feature = "zstd")]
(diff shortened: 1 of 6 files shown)


src/layers/tile/util.rs
```diff
@@ -11,47 +11,28 @@ pub(crate) fn parse_data_line(
...
-                encoding,
-                compression,
-            })
-        }
-    };
+        (Some("csv"), None) => decode_csv(parser, tilesets),
+
+        (Some("base64"), None) => parse_base64(parser).map(|v| convert_to_tiles(&v, tilesets)),
+        (Some("base64"), Some("zlib")) => parse_base64(parser)
+            .map(|data| std::io::Cursor::new(data))
```
bjorn, line 18: I still don't really like this duplication of `parse_base64` and `convert_to_tiles` for each base64-encoded data, but in trying to make it shorter we of course get duplicated error handling and I had to make a clone of the encoding and compression for the error:
```rust
    match (encoding.as_deref(), compression.as_deref()) {
        (Some("csv"), None) => decode_csv(parser, tilesets),
        (Some("base64"), c) => {
            parse_base64(parser).and_then(|bytes| {
                match c {
                    None => Ok(bytes),
                    Some("zlib") => process_decoder(libflate::zlib::Decoder::new(&bytes[..])),
                    Some("gzip") => process_decoder(libflate::gzip::Decoder::new(&bytes[..])),
                    #[cfg(feature = "zstd")]
                    Some("zstd") => process_decoder(zstd::stream::read::Decoder::with_buffer(&bytes[..])),
                    _ => Err(TiledError::InvalidEncodingFormat {
                        encoding: encoding.clone(),
                        compression: compression.clone(),
                    }),
                }
            }).and_then(|v| Ok(convert_to_tiles(&v, tilesets)))
        },
        _ => Err(TiledError::InvalidEncodingFormat {
            encoding,
            compression,
        }),
    }
```

So, I wouldn't suggest to go with the above.

However, it seems we can at least get rid of the `std::io::Cursor`, since we can just do:
```rust
|data| process_decoder(libflate::zlib::Decoder::new(&data[..]))
```
Or is there some disadvantage to doing that?

bjorn (commented):
Nice overall simplification! :-)

src/layers/tile/util.rs
alexdevteam: Didn't know `Read` was implemented for `&[u8]`, will change later

And yes, that base64 duplication is annoying but as far as i know there isn't any clean way of changing it

bjorn (approved):

PR #156, opened 2022-02-12
Changed Rc<Tileset> to Arc<Tileset>

Anti-Alias:
Hey, thanks for accepting my PR a few days ago.
It was my first contribution to an open source project!

Right now, I'm working on a project using the Bevy game engine and am making an AssetLoader for TiledMap.
Bevy is inherently multithreaded by design, and Assets can be shared across Threads, so they must implement Send + Sync.

I'm still new to Rust, and I found out the hard way that all structs implement Send + Sync automatically unless they have fields that do not implement Send + Sync.

Rc is one of those types.

It's for this reason that I'd like to change Rc<Tileset> to Arc<Tileset> as this would allow for a Map to be sent and shared between threads. I don't believe that this would impact performance much since the reference count for tilesets is really only changed during the creation and destruction of the map as far as I can tell. I'm not sure if this is subject to change.

I'd like to know what you guys think, though.

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -17,6 +17,9 @@ jobs:
     - uses: actions/checkout@v2
     - uses: Swatinem/rust-cache@v1
 
+    - name: Update package manager
+      run: sudo apt-get update
+
     - name: Install dependencies
       run: sudo apt-get install -y libsfml-dev libcsfml-dev
 
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -15,11 +15,13 @@ jobs:
 
     steps:
     - uses: actions/checkout@v2
-
     - uses: Swatinem/rust-cache@v1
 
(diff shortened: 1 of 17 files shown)


alexdevteam (approved):
Looks good, thanks!

alexdevteam:
We'll wait until the #154 hotfix merges first, though.

alexdevteam:
@Anti-Alias Could you merge with master?

bjorn (approved):
Yeah, I guess it's not worth making this somehow optional. Indeed there should not be any noticeable impact on performance in making the reference counting atomic.

Anti-Alias:
@alexdevteam I don't seem to have write access, so I cannot merge myself.

Anti-Alias:
> 

Aha, didn't realize.
Sure, I'll do that now.

alexdevteam:
Thank you for your contribution!

PR #157, opened 2022-02-12
Refactor chunk system

alexdevteam:
- Add `FiniteTileLayer` & `InfiniteTileLayer`
- Store chunks by chunk position
- Add `InternalChunk` for holding mid-parse chunk info

### DIFF
--- a/examples/main.rs
+++ b/examples/main.rs
@@ -17,22 +17,25 @@ fn main() {
         print!("Layer \"{}\":\n\t", layer.data().name);
 
         match layer.layer_type() {
-            tiled::LayerType::TileLayer(layer) => match layer.data() {
-                tiled::TileLayerData::Finite(data) => println!(
+            tiled::LayerType::TileLayer(layer) => match layer {
+                tiled::TileLayer::Finite(data) => println!(
                     "Finite tile layer with width = {} and height = {}; ID of tile @ (0,0): {}",
-                    data.width(),
-                    data.height(),
-                    layer.get_tile(0, 0).unwrap().id
+                    data.data().width(),
+                    data.data().height(),
+                    data.get_tile(0, 0).unwrap().id
                 ),
-                tiled::TileLayerData::Infinite(data) => {
+                tiled::TileLayer::Infinite(data) => {
                     // This is prone to change! Infinite layers will be refactored before 0.10.0
                     // releases.
-                    println!("Infinite tile layer with {} chunks", data.chunks.len())
+                    println!(
+                        "Infinite tile layer; Tile @ (-5, 0) = {:?}",
+                        data.get_tile(-5, 0)
+                    )
                 }
             },
             tiled::LayerType::ObjectLayer(layer) => {
                 println!("Object layer with {} objects", layer.data().objects.len())
-            },
+            }
             tiled::LayerType::ImageLayer(layer) => {
                 println!(
                     "Image layer with {}",
@@ -42,12 +45,9 @@ fn main() {
                         None => "no image".to_owned(),
                     }
                 )
-            },
+            }
             tiled::LayerType::GroupLayer(layer) => {
-                println!(
-                    "Group layer with {} sublayers",
-                    layer.layers().len()
-                )
+                println!("Group layer with {} sublayers", layer.layers().len())
             }
         }
     }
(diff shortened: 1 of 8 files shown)


alexdevteam:
Probably filled with bugs. Still need to add tests and fix the already existing one.

alexdevteam:
Turns out it was the test which was wrong, not the implementation. Phew

examples/main.rs
```diff
@@ -17,22 +17,25 @@ fn main() {
...
+                tiled::TileLayer::Infinite(data) => {
                     // This is prone to change! Infinite layers will be refactored before 0.10.0
                     // releases.
-                    println!("Infinite tile layer with {} chunks", data.chunks.len())
+                    println!(
+                        "Infinite tile layer; Tile @ (-5, 0) = {:?}",
+                        data.get_tile(-5, 0)
+                    )
                 }
             },
```
bjorn, line 35: If we're removing apparently unnecessary commas here, let's also remove this one. :-)

src/layers/tile/finite.rs
```diff
@@ -51,9 +51,9 @@ impl FiniteTileLayerData {
         })
     }
 
-    pub(crate) fn get_tile(&self, x: usize, y: usize) -> Option<&LayerTileData> {
-        if x < self.width as usize && y < self.height as usize {
-            self.tiles[x + y * self.width as usize].as_ref()
+    pub(crate) fn get_tile(&self, x: i32, y: i32) -> Option<&LayerTileData> {
+        if x < self.width as i32 && y < self.height as i32 {
```
bjorn, line 55: I guess we should also check that these values are at least 0 now.

src/layers/tile/infinite.rs
```diff
@@ -39,33 +39,81 @@ impl InfiniteTileLayerData {
...
     tiles: Vec<Option<LayerTileData>>,
 }
 
 impl Chunk {
+    pub const WIDTH: u32 = 16;
+    pub const HEIGHT: u32 = 16;
+
+    pub(crate) fn new() -> Self {
+        Self {
+            tiles: vec![None; Self::WIDTH as usize * Self::HEIGHT as usize],
```
bjorn, line 96: If we use static size, would it be better to use an array instead of a `Vec`, or is that not helpful regarding moving it around in memory as part of its use in a hash map?

src/layers/tile/infinite.rs
```diff
@@ -39,33 +39,81 @@ impl InfiniteTileLayerData {
...
+        }
+    }
+}
+
+#[derive(Debug, PartialEq, Clone)]
+struct InternalChunk {
+    first_tile_pos: (i32, i32),
```
bjorn, line 103: Personally I'd just call this `x` and `y` of the chunk, as in the TMX format.

bjorn (approved):
Very nice to see infinite layers supported! I left a few comments.

examples/main.rs
alexdevteam: If I accidentally removed a comma, that was probably done by rustfmt; Details like that aren't really that important and don't create any issues for the parser/compiler in general.

src/layers/tile/infinite.rs
alexdevteam: True; We could wrap it in a `Box` to ensure moves aren't expensive.

src/layers/tile/infinite.rs
alexdevteam: I avoided that naming because it is inprecise; While it could refer to the tmx format it could also refer to the chunk's position in chunk units. This name avoids that confusion.

bjorn: Hmm, well if the position could be in chunk units, then so could be its width and height... :-)

src/layers/tile/finite.rs
```diff
@@ -51,9 +51,9 @@ impl FiniteTileLayerData {
         })
     }
 
-    pub(crate) fn get_tile(&self, x: usize, y: usize) -> Option<&LayerTileData> {
-        if x < self.width as usize && y < self.height as usize {
-            self.tiles[x + y * self.width as usize].as_ref()
+    pub(crate) fn get_tile(&self, x: i32, y: i32) -> Option<&LayerTileData> {
+        if x < self.width as i32 && y < self.height as i32 && x >= 0 && y >= 0 {
+            self.tiles[(x + y * self.width as i32) as usize].as_ref()
```
bjorn, line 56: I think it would be better to convert to `usize` before multiplying, since the i32 has a narrower range. So this should probably be something like:
```suggestion
            self.tiles[x as usize + y as usize * self.width].as_ref()
```

bjorn (approved):

bjorn (approved):

Issue #134, opened 2022-01-26
Better / More examples


alexdevteam:
The [only example currently](https://github.com/mapeditor/rs-tiled/blob/master/examples/main.rs) has no notion about how to use the library. There should be a more involved example available that doesn't necessarily use a renderer (Such as #133) but still uses most, if not all, of the functionality present in the library.

PieKing1215:
I'm working on an example like #133 for ggez as well

aleokdev:
The last example I'd like to fit in would be a `bevy` one, which would entirely showcase the VFS system (And will most definitely discover new interface issues we'll have to tackle)

That should be it. If the crate's good for bevy, ggez and SFML, then it should be viable with any other game engine crate, given the difference between all of the ones we'll have.

Issue #158, opened 2022-02-12
Remove `Data` types from the interface


alexdevteam:
#135 forced us to think about some way to replace GIDs with something else that would uniquely identify a tile. References were out of the question since lifetimes would have been a pain to deal with; As such, we ended up on a design which divided types into two categories: `Data` and `MapWrapper<'map, Data>`. The first one stores all the raw data related to the object and fully owns it; The second one has a reference to the data and its parent map. This design was chosen because it would allow to have layer functions which relied in map information without having the user pass the map into these functions.

The issue now is that, although this division makes sense internally, user-side it is a mess. Functions that require the map are put in the wrapper, while everything else is in the `Data` member accessed through `.data()`. Sometimes the usage of `.data()` feels somewhat random; it is also inconsistent for some types such as `LayerType`, which has no data counterpart and is actually not a `MapWrapper` specialization.
As of #157 the interface looks like this:
![Screenshot_20220212_160822](https://user-images.githubusercontent.com/16354904/153716706-1b75ce11-a147-4634-ab93-0211ca1b7f5c.png)

**I suggest to make Data types internal, and replace them in the interface with properties in their respective wrappers.** The main disadvantage of this approach is that it adds a lot of boilerplate code to the library (one function per data member) but it has a lot of advantages such as being able to control more precisely what the user gets access to, and having a much simpler interface.

What do you think?

bjorn:
Thanks for the nice visual overview! Your proposal makes a lot of sense to me. It'd be great if we can get rid of the `.data()` calls this way.

In C++ it is common practice to keep the internal data structure private and only accessible through member functions, in the interest of keeping binary compatibility. I guess this kind of compatibility is also useful in Rust, maybe even more so because adding members can even break source compatibility.

PR #161, opened 2022-02-14
Add ggez example

PieKing1215:
Adds a rendering demo for [ggez](https://github.com/ggez/ggez).
Supports multiple layers, multiple tilesets, and parallax.
Also draws basic shape objects.

You can middle mouse + drag to pan around and scroll to zoom.
Right click also toggles a demo animation that makes the tiles move around.

It has pretty good performance because it uses ggez's `SpriteBatch`es for tile rendering. It could be better by caching and reusing the batches but I figured I'd leave that out to simplify a bit.

I made a nicer example map for testing locally but I'll make a separate PR for that I think.
Since the current example map is large and mostly empty, and this centers it in the window at the start, you need to zoom out or pan in order to see it (the other nicer map has a more reasonable size).

(this does not properly hook into ggez's filesystem for loading external tilesets/images as that would still require something like #100 I believe)

This could probably still use a bit more clean up and commenting

related: #134


bjorn:
It did not compile for me:
```
error: there is no argument named `fps`
  --> examples/ggez/main.rs:83:49
   |
83 |         let text = graphics::Text::new(format!("{fps:.0} fps"));
   |                                                 ^^^^^^^^

error: could not compile `tiled` due to previous error
```

PieKing1215:
> error: there is no argument named `fps`

I just changed it so it should compile on older rust versions (that feature was only just added in rust 1.58)

examples/ggez/main.rs
```diff
@@ -0,0 +1,130 @@
...
+    // init ggez
+    let cb = ggez::ContextBuilder::new("rs-tiled + ggez", "rs-tiled")
+        .window_setup(ggez::conf::WindowSetup::default()
+            .title("rs-tiled + ggez example")
+            .vsync(false)
+        )
+        .window_mode(ggez::conf::WindowMode::default()
+            .dimensions(1000.0, 800.0)
+        )
+        .add_resource_path(""); // add repo root to ggez filesystem (our example map looks for `assets/tilesheet.png`)
```
alexdevteam, line 17: I believe this should be `std::env::var("CARGO_MANIFEST_DIR")` so that the example works in directories within the root one

examples/ggez/map.rs
```diff
@@ -0,0 +1,239 @@
...
+
+    pub fn background_color(&self) -> Option<graphics::Color> {
+        self.map.background_color
+            .map(|c| ggez::graphics::Color::from_rgba(c.red, c.green, c.blue, c.alpha))
+    }
+
+    pub fn draw(&mut self, ctx: &mut Context, draw_param: DrawParam, parallax_pan: (f32, f32)) -> GameResult {
+
+        // could be cached for more performance
+        let layer_batches: Vec<Vec<SpriteBatch>> = self.generate_map_render(ctx, parallax_pan);
```
alexdevteam, line 70: Yeah this should definitely be cached

alexdevteam (commented):
I'd appreciate if you add an example description as doc comments to the main file (like in the SFML example). Apart from that I've got two comments:

examples/ggez/map.rs
```diff
@@ -0,0 +1,239 @@
...
+
+        layer_batches
+    }
+    
+    fn draw_object(object: &tiled::ObjectData, ctx: &mut Context, draw_param: DrawParam) -> GameResult {
+        match &object.shape {
+            tiled::ObjectShape::Rect { width, height } => {
+                let bounds = graphics::Rect::new(object.x, object.y, *width, *height);
+                let shape =
+                    graphics::Mesh::new_rectangle(ctx, graphics::DrawMode::stroke(2.0), bounds, graphics::Color::CYAN)?;
```
bjorn, line 181: Note that there is also `Object::get_tile`, which would be nice to support as well. Of course then the object needs to be accessed through `ObjectLayer::get_object` instead of iterating `ObjectLayerData::objects` directly.

Since this might be a bit of a cludge, we could wait with supporting that until #158 is resolved.

bjorn (commented):
Great and smooth running example! It showed over 5000 FPS for me on my aging GTX 1060, and over 7000 FPS when caching the `layer_batches`. Nice animation too! I've just left some minor comments.

examples/ggez/map.rs
```diff
@@ -0,0 +1,239 @@
...
+                                // get tile's rectangle in the tileset texture
+                                let rect = get_tile_rect(tile.tileset, tile.id);
+                                let ts = tile.tileset;
+                                if let Some((batch, ts_size)) = ts_sizes_and_batches.get_mut(&ts.name) {
+                                    let mut dx = x as f32 * self.map.tile_width as f32 + parallax_pan.0 * (layer.parallax_x - 1.0);
+                                    let mut dy = y as f32 * self.map.tile_height as f32 + parallax_pan.1 * (layer.parallax_y - 1.0);
+
+                                    if self.example_animate {
+                                        dx += (ggez::timer::time_since_start(ctx).as_secs_f32() - x as f32 * 0.3 + i as f32 * 0.25).sin() * 20.0;
+                                        dy += (ggez::timer::time_since_start(ctx).as_secs_f32() * 1.25 + y as f32 * 0.3 + i as f32 * 0.25).cos() * 20.0;
```
alexdevteam, line 138: You should probably put `ggez::timer::time_since_start(ctx).as_secs_f32()` in a let-binding so as to not repeat it and call it only once

examples/ggez/map.rs
```diff
@@ -0,0 +1,239 @@
...
+        for l in self.map.layers() {
+            match &l.layer_type() {
+                tiled::LayerType::ObjectLayer(ol) => {
+                    for o in &ol.data().objects {
+                        Self::draw_object(o, ctx, draw_param.clone())?;
+                    }
+                }
+                _ => {},
+            }
+        }
```
alexdevteam, line 93: I don't get it, why are these separate? We should render the layers in order, shouldn't we?

examples/ggez/main.rs
```diff
@@ -0,0 +1,130 @@
...
+        let text = graphics::Text::new(format!("{:.0} fps", fps));
+
+        graphics::draw(
+            ctx,
+            &text,
+            DrawParam::default()
+                .dest([window_size.0 - text.width(ctx) - 40.0, 10.0])
+                .scale([1.25, 1.25])
+                .color(graphics::Color::WHITE),
+        )?;
```
alexdevteam, line 92: You can replace these comments followed by code by functions; The end result will be cleaner and easier to follow

examples/ggez/map.rs
```diff
@@ -0,0 +1,239 @@
...
+                                        dy += (ggez::timer::time_since_start(ctx).as_secs_f32() * 1.25 + y as f32 * 0.3 + i as f32 * 0.25).cos() * 20.0;
+                                    }
+
+                                    batch.add(
+                                        DrawParam::default()
+                                            .src(ggez::graphics::Rect::new(
+                                                rect.0 as f32 / (*ts_size).0 as f32,
+                                                rect.1 as f32 / (*ts_size).1 as f32,
+                                                rect.2 as f32 / (*ts_size).0 as f32,
+                                                rect.3 as f32 / (*ts_size).1 as f32,
```
alexdevteam, line 147: Wouldn't it be easier for `get_tile_rect` to return the graphics rect directly?

PieKing1215: Yeah probably, I don't remember if there was any particular reason I didn't do that but I'll change it when I get around to updating this

examples/ggez/main.rs
alexdevteam:
```suggestion
        // add repo root to ggez filesystem (our example map looks for `assets/tilesheet.png`)
        .add_resource_path("std::env::var("CARGO_MANIFEST_DIR").unwrap()");
```

examples/ggez/map.rs
alexdevteam: Here, I have the function from another project:
```rs
fn get_tile_rect(
    tileset: &tiled::Tileset,
    id: u32,
    ts_img_width: u16,
    ts_img_height: u16,
) -> graphics::Rect {
    let ts_x = id % tileset.columns;
    let ts_y = id / tileset.columns;

    let x = (tileset.margin + (tileset.tile_width + tileset.spacing) * ts_x) as f32;
    let y = (tileset.margin + (tileset.tile_height + tileset.spacing) * ts_y) as f32;

    let ts_img_width = ts_img_width as f32;
    let ts_img_height = ts_img_height as f32;

    graphics::Rect {
        x: x / ts_img_width,
        y: y / ts_img_height,
        w: tileset.tile_width as f32 / ts_img_width,
        h: tileset.tile_height as f32 / ts_img_height,
    }
}
```

PieKing1215:
Cleaned up some of the easy stuff, still some organization and structural things left to do, and I still need to merge master into here again

examples/ggez/map.rs
```diff
@@ -0,0 +1,249 @@
...
+    pub example_animate: bool,
+}
+
+impl MapHandler {
+    pub fn new(map: tiled::Map, ctx: &mut Context) -> GameResult<Self> {
+        // load images for the map's tilesets
+        let mut tileset_image_cache = HashMap::new();
+        for ts in map.tilesets().iter() {
+            if let Some(image) = &ts.image {
+                // image path comes in as "assets/tilesheet.png"
```
aleokdev, line 18: Is this true? I thought image sources were absolute; We'll need to fix this

examples/ggez/map.rs
```diff
@@ -0,0 +1,249 @@
...
+
+        // draw tile layers
+
+        // for each layer
+        for layer in &layer_batches {
+            // for each tileset in the layer
+            for batch in layer {
+                graphics::draw(ctx, batch, draw_param)?;
+            }
+        }
```
aleokdev, line 80: Could be flattened to a single vector if layer batches are sequentially placed by index

bjorn:
I've merged `master` and fixed the compile, and also adjusted `main.rs` to use the new `Loader` class. However, for me this example is currently just displaying a magic pink window. Any idea why I'm no longer seeing the map?

@PieKing1215 Do you think you could find some time to finalize this example? I think it'd be really nice to have this merged, and it would eventually also be able to demonstrate the VFS support that has been merged for 0.11 (#199).

bjorn:
@aleokdev It appears the clippy check needs also the dependencies installed. Maybe it would make sense to merge it with the build check?

aleokdev:
The CI deps can be added in this PR as they're needed by this example. Is that what you're referring to? I'm not sure if I understood you.

bjorn:
> The CI deps can be added in this PR as they're needed by this example. Is that what you're referring to? I'm not sure if I understood you.

I just mean that it's something annoying if we need to install the dependencies in two places in the workflow, just because the build and clippy checks happen on separate virtual machines. It might be better to just perform those checks at two separate steps in on the same machine.

aleokdev:
Oh, sure, that makes sense.

bjorn:
> Oh, sure, that makes sense. However, I'd like for them to be treated as separate checks so that we can see at a glance what's wrong. Is that possible?

I agree that would be nice, but I have no idea whether this is possible. Maybe it can be done with [this github-checks action](https://github.com/marketplace/actions/github-checks).

PieKing1215:
> @PieKing1215 Do you think you could find some time to finalize this example? I think it'd be really nice to have this merged, and it would eventually also be able to demonstrate the VFS support that has been merged for 0.11 (#199).

Yeah sorry about the lack of progress on this, I've been particularly busy recently but I should be able to dedicate some time to adding the rest of the suggested things.
(Thanks for merging master btw)

> However, for me this example is currently just displaying a magic pink window. Any idea why I'm no longer seeing the map?

Did you try zooming out? It centers the map at the start but the test map's size is way too big and mostly empty so you can't see anything without zooming/panning (should probably fix this)

bjorn:
> Did you try zooming out? It centers the map at the start but the test map's size is way too big and mostly empty so you can't see anything without zooming/panning (should probably fix this)

Heh, it had been a while since I had run this example, and I had just forgotten about the need to zoom out, whoops! But yes, that should definitely be fixed.

examples/ggez/map.rs
PieKing1215: I added this cache and a function to invalidate it, which should hopefully make it obvious why/how to use it.
It calls invalidate when you pan/zoom or every frame while animating

PieKing1215:
The main things I can think of that are unsupported in this right now are:
- Tile objects
- Infinite maps
- ImageLayer
- GroupLayer

I plan on doing tile objects already, but I'm not sure about the other stuff. Idk how all-encompassing we want this example to be.
Also should I do VFS in this PR (and change base branch to 0.11) or leave that for the future?

bjorn:
> I plan on doing tile objects already, but I'm not sure about the other stuff. Idk how all-encompassing we want this example to be.

I think tile objects would be good to cover, but the other stuff is not so important.

> Also should I do VFS in this PR (and change base branch to 0.11) or leave that for the future?

I would suggest to target 0.10 first and update the example after merging it to 0.11, unless the lack of using VFS is really setting a bad example and we'd rather want to wait till 0.11 to show how to use this crate with ggez at all.

aleokdev:
I'd wait until 0.11 to be honest, VFS is a huge deal with ggez (Plus 0.11's interface is basically finished at this point)

aleokdev:
Should we base this on 0.11 then?

bjorn:
@aleokdev It's fine with me too.

aleokdev:
@PieKing1215 I'm basing this onto 0.11 then; You'll be able to use custom loaders now. My platformer template repo has an example of their usage:

```rs
// Need to do newtype to implement ResourceReader for ggez's filesystem
// FIXME: This would greatly improve with separated subcontexts (ggez 0.8.0)
pub struct FsContext<'ctx>(pub &'ctx mut ggez::Context);

impl tiled::ResourceReader for FsContext<'_> {
    type Resource = filesystem::File;

    type Error = GameError;

    fn read_from(
        &mut self,
        path: &std::path::Path,
    ) -> std::result::Result<Self::Resource, Self::Error> {
        filesystem::open(&self.0, path)
    }
}
```
Then you can simply
```rs
let mut loader = tiled::Loader::with_cache_and_reader(
     tiled::DefaultResourceCache::new(),
     FsContext(ctx),
);
```
And not worry about paths anymore. When tiled needs to load any supplementary data, it'll use ggez's filesystem to do so.

bjorn (approved):
The map could be nicer, but I think the example is good to go now.

PR #163, opened 2022-02-16
Remove `Data` types from the interface

### LINKED ISSUE #158: Remove `Data` types from the interface

alexdevteam:
Closes #162 & closes #158.

### DIFF
--- a/src/layers/group.rs
+++ b/src/layers/group.rs
@@ -85,10 +85,10 @@ pub type GroupLayer<'map> = MapWrapper<'map, GroupLayerData>;
 
 impl<'map> GroupLayer<'map> {
     pub fn layers(&self) -> GroupLayerIter {
-        GroupLayerIter::new(self.map(), self.data())
+        GroupLayerIter::new(self.map, self.data)
     }
     pub fn get_layer(&self, index: usize) -> Option<Layer> {
-        self.data().layers.get(index).map(|data| Layer::new(self.map(), data))
+        self.data.layers.get(index).map(|data| Layer::new(self.map, data))
     }
 }
 
--- a/src/layers/group.rs
+++ b/src/layers/group.rs
@@ -11,7 +11,7 @@ use crate::{
 };
 
 #[derive(Debug, PartialEq, Clone)]
-pub struct GroupLayerData {
+pub(crate) struct GroupLayerData {
     layers: Vec<LayerData>,
 }
 
--- a/src/layers/group.rs
+++ b/src/layers/group.rs
@@ -1,13 +1,13 @@
-use std::path::Path;
 use std::collections::HashMap;
+use std::path::Path;
 
-use crate:: {
-    layers::{LayerData, LayerTag},
+use crate::{
     error::TiledError,
-    properties::{parse_properties, Properties},
+    layers::{LayerData, LayerTag},
     map::MapTilesetGid,
+    properties::{parse_properties, Properties},
     util::*,
-    MapWrapper, Layer, Map
+    Layer, Map,
 };
 
 #[derive(Debug, PartialEq, Clone)]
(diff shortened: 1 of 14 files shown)


alexdevteam:
I'd love to use a proc macro instead of `map_wrapper` so we wouldn't have duplicate docs in data members + interface properties but both `syn` and `quote` are really heavy to use.

alexdevteam:
`Tile::collision` is not implemented yet. Having an object group there is giving a few issues.

alexdevteam:
Had to end up making both `ObjectLayerData` and `ObjectData` public. There isn't really any issue with that and it's probably the type collisions should use (It's raw object data, not belonging to any map). However it's inconsistent with everything else.

bjorn:
> There isn't really any issue with that and it's probably the type collisions should use (It's raw object data, not belonging to any map). However it's inconsistent with everything else.

Yeah, the inconsistency isn't very nice, and in fact I've been struggling with the same in Tiled, for example to avoid placing tile objects in these object groups and workarounds for assigning IDs to these objects, which are normally counted per map. But I guess we can live with it.

src/layers/group.rs
```diff
@@ -101,7 +101,11 @@ pub struct GroupLayerIter<'map> {
 
 impl<'map> GroupLayerIter<'map> {
     fn new(map: &'map Map, group: &'map GroupLayerData) -> Self {
-        Self { map, group, index: 0 }
+        Self {
+            map,
+            group,
+            index: 0,
+        }
```
bjorn, line 108: What is the logic behind these reformattings? Here it's wrapping them all on their own line, but just a few lines up you removed some wrapping to form `Ok((Self { layers }, properties))`. I could be fine with either notation, but why keep changing these around?

src/layers/mod.rs
```diff
@@ -112,12 +110,62 @@ impl LayerData {
...
+    pub fn offset_x(&self) -> f32 {
+        self.data.offset_x
+    }
+
+    /// Get the layer's y offset (in pixels).
+    pub fn offset_y(&self) -> f32 {
+        self.data.offset_y
+    }
+
+    /// Get the layer's x parallax.
```
bjorn, line 141:
```suggestion
    /// Get the layer's x parallax factor.
```

src/layers/mod.rs
```diff
@@ -112,12 +110,62 @@ impl LayerData {
...
+    pub fn parallax_x(&self) -> f32 {
+        self.data.parallax_x
+    }
+
+    /// Get the layer's y parallax.
```
bjorn, line 146:
```suggestion
    /// Get the layer's y parallax factor.
```

bjorn (approved):
Great work!

bjorn (approved):

src/layers/group.rs
alexdevteam: No reason for it, just rustfmt doing its thing.

bjorn: Can we do a single pass of rustfmt to avoid these changes, or does it actually like to change the wrapping back and forth?

Issue #145, opened 2022-02-07
Support for templates


aleokdev:
Support for object templates

bjorn:
Closing issue since it is done now that #170 was merged. Thanks for your help, @RossBrunton!

PR #170, opened 2022-02-22
Add template support

RossBrunton:
Whew, this was a bigger change than I thought it would be. I've made a few assumptions about the best way to design and implement things, so feel free to let me know if it should be restructured any.

Relates to #145

======

This patch adds support for templates. Templates are a property on Objects
that specify an external file to load object properties by default. In
this implementation, templates are not visible to the end user, their
behaviour happens "automatically".
    
As it parses, the Map will generate a list of templates, with each being
identified by their ID.
    
Templates introduce the concept of a tile that is bound to a tileset
that is not part of the main map's list of tilesets. For this reason,
the tileset_index() function now returns Option<usize>. Internally, an
enum is used to signify whether the tileset is map or template bound.
    
I had to change the signature of the Resource cache, because the
generator function now needs access to the cache.


bjorn:
This change really makes me wonder whether I made some horrible design decisions, and whether it couldn't all be a lot simpler when we could make a few changes to the data format. If you have any ideas about this, please let me know so we could potentially simplify this in future versions of Tiled.

Unfortunately, I won't have time to give this a thorough review until next week.

One thing sticks out to me, which is the choice to make the templates invisible to the user. That sounds very convenient, but eventually I still have the idea that `rs-tiled` could also be used to make modifications and write back files. Will it be possible to add this in the future without major changes in the API? Of course @aleokdev has been telling me not to worry about modification and writing for now (and maybe by then we should reconsider the entire format), but I just thought I'd mention it.

aleokdev:
@bjorn If you ever do consider a new format, please count me in in the new specs team. :)

src/template.rs
```diff
@@ -0,0 +1,142 @@
...
+            return Ok(id);
+        }
+
+        // Open the template file
+        let file = File::open(&path).map_err(|err| TiledError::CouldNotOpenFile {
+            path: path.to_path_buf(),
+            err,
+        })?;
+
+        let mut tileset_parser = EventReader::new(file);
```
aleokdev, line 48: *Tileset* parser?

src/template.rs
```diff
@@ -0,0 +1,142 @@
...
+        parse_context: &mut ParseContext<CacheTy>,
+    ) -> Result<usize, TiledError> {
+        let mut object = Option::None;
+
+        let template_id = parse_context.templates.len();
+        parse_context.templates.push(Template {
+            path: path.to_path_buf(),
+            tileset: None,
+            object: None,
+        });
```
aleokdev, line 87: Again, this shouldn't be here; These are changes you wouldn't expect from a function called `parse_external_template`.

aleokdev (changes requested):
IMO `ParseContext` should not be defined in this PR, as it is largely unrelated to templates. Yes, it simplifies things, but it also made reviewing the PR much harder, and it has many issues unrelated to the addition of templates. I'd suggest reverting these changes; I'd be happy to review a parse context PR after this one has been merged so we can get the big picture.

RossBrunton:
> IMO `ParseContext` should not be defined in this PR, as it is largely unrelated to templates. Yes, it simplifies things, but it also made reviewing the PR much harder, and it has many issues unrelated to the addition of templates. I'd suggest reverting these changes; I'd be happy to review a parse context PR after this one has been merged so we can get the big picture.

I didn't really want to add it here as well, but we need to pass a lot more information around, and to a lot of locations. Specifically to parse templates we need:
* The cache (to load a tilesheet if it's been already been processed).
* The base filename (for the base file that the tile URI is based on).

Templates can be specified any time we need to process an Object, which comes up in a lot of places (although I guess we could just not support templates in more obscure places, like tilesets).

Templates also introduce a new "context" which affects which tileset a tile is linked to (not the ones in the main map), which is another piece of information that needs to be propogated up and down the call stack.

I'm open to ideas on how we can improve it, but I can't see any solution that doesn't involve giving each function 2-3 more arguments. Using a struct to store this information just makes it more cleaner IMO. I can refactor things so that arguments are passed directly between functions if that is what you prefer.

src/template.rs
```diff
@@ -0,0 +1,142 @@
...
+    pub path: PathBuf,
+    pub tileset: Option<Arc<Tileset>>,
+    pub object: Option<ObjectData>,
+}
+
+impl Template {
+    pub(crate) fn new_external<CacheTy: ResourceCache>(
+        path: &Path,
+        parse_context: &mut ParseContext<CacheTy>,
+    ) -> Result<usize, TiledError> {
```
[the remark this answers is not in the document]
RossBrunton: So there's an issue here in that tiles contained in a Template need to be able to store a reference of some kind to the template (so they can identify their tileset). I could have the Tile contain an Arc to either the Tileset or the Template, but that would make the Tile not be Copy, and may have performance implications.

I decided to copy how it was done with Tilesets, where during parsing a list of tilesets is generated, and tiles store an index into the list. For templates though, the parsing of the template's tile is done during parsing of the template itself, so the template needs to know its own id to tell the tile.

I agree this is a bit of a mess though, and will gladly change it if there is a better way. And yeah, the function names are not idiomatic, sorry!

RossBrunton:
> This change really makes me wonder whether I made some horrible design decisions, and whether it couldn't all be a lot simpler when we could make a few changes to the data format. If you have any ideas about this, please let me know so we could potentially simplify this in future versions of Tiled.
> 
> Unfortunately, I won't have time to give this a thorough review until next week.
> 
> One thing sticks out to me, which is the choice to make the templates invisible to the user. That sounds very convenient, but eventually I still have the idea that `rs-tiled` could also be used to make modifications and write back files. Will it be possible to add this in the future without major changes in the API? Of course @aleokdev has been telling me not to worry about modification and writing for now (and maybe by then we should reconsider the entire format), but I just thought I'd mention it.

From my (limited) experience of working with Tiled files, I think the format is pretty solid. I've found templates a really useful addition for managing object properties. The complexity with this MR is more with managing file locations and in-process parsing, neither of which I think are indicative of problems in the file format.

With regards to the templates being resolved "automatically". It should be simple enough to change things so that objects store a template index number, and properties (x, y, width, etc.) on ObjectData being optional (which is more faithful to the original data in the file anyway). The getter methods would check for the data on the ObjectData, then the template, and finally a default value.

I think if the library is automatically decoding and calculating gids to tileset indexes, it makes sense for it to also automatically parse and understand templates as well.

aleokdev:
> but I can't see any solution that doesn't involve giving each function 2-3 more arguments

It's okay, we'll deal with those later. What is important right now is the feature; refactors can come later down the line, which also allows us to think more about them as we have more information about the interface at our disposal.

bjorn:
> From my (limited) experience of working with Tiled files, I think the format is pretty solid. I've found templates a really useful addition for managing object properties. The complexity with this MR is more with managing file locations and in-process parsing, neither of which I think are indicative of problems in the file format.

Alright, that's great to hear!

> With regards to the templates being resolved "automatically". It should be simple enough to change things so that objects store a template index number, and properties (x, y, width, etc.) on ObjectData being optional (which is more faithful to the original data in the file anyway). The getter methods would check for the data on the ObjectData, then the template, and finally a default value.

Rather than making each value optional, in Tiled there is a class member with bitflags indicating which values have been "overridden". In addition to being less fragmented, this also allows copying down values from the template so access can be more direct. Of course that does mean that when editing the template, the changes need to be copied down to all instances. Anyway, we can worry about that later, when we get around to #171.

> I think if the library is automatically decoding and calculating gids to tileset indexes, it makes sense for it to also automatically parse and understand templates as well.

Yes, that is definitely helpful.

src/map.rs
```diff
@@ -174,15 +181,23 @@ impl Map {
...
-                        let file = File::open(&tileset_path).map_err(|err| TiledError::CouldNotOpenFile{path: tileset_path.clone(), err })?;
-                        let tileset = cache.get_or_try_insert_tileset_with(tileset_path.clone(), || Tileset::new_external(file, Some(&tileset_path)))?;
+                        let tileset = if let Some(ts) = cache.get_tileset(&tileset_path) {
+                            ts
+                        } else {
+                            let file = File::open(&tileset_path).map_err(|err| TiledError::CouldNotOpenFile{path: tileset_path.clone(), err })?;
+                            let tileset = Arc::new(Tileset::new_external(file, Some(&tileset_path), &tilesets, &mut templates, None, cache)?);
+                            cache.insert_tileset(tileset_path.clone(), tileset.clone());
+                            tileset
+                        };
```
aleokdev, line 200: This change isn't related to templates; I suggest we keep the original `get_or_try` method.

src/objects.rs
```diff
@@ -41,27 +42,70 @@ impl ObjectData {
...
+
+        // If the template attribute is there, we need to go fetch the template file
+        let template_id = if let Some(template) = template {
+            let s: String = template;
+            let parent_dir = base_path
+                .ok_or(TiledError::SourceRequired {
+                    object_to_parse: "Template".to_string(),
+                })?
+                .parent()
+                .unwrap();
```
aleokdev, line 82: This should return an error instead; I believe there was an `InvalidPath` error or similar in the interface.

src/tileset.rs
```diff
@@ -113,10 +136,23 @@ impl Tileset {
     pub(crate) fn parse_xml_in_map(
         parser: &mut impl Iterator<Item = XmlEventResult>,
         attrs: Vec<OwnedAttribute>,
-        map_path: &Path,
+        map_path: &Path, // Template or Map file
```
aleokdev, line 139: Uuuuuuhm...

aleokdev (changes requested):
Thank you! This is definitely much better. However, there are still a couple of important issues that need to be tackled.

aleokdev:
BTW, you don't need to force push on each commit; In fact, it's better for us that you keep the original commits so we can see more easily what's changed between each review. Don't worry, the commits will all be squeezed down to a single one when merging with master, so the PR history can be as messy as you want (Still, be a bit careful :) ) Thanks!

aleokdev:
> BTW, you don't need to force push on each commit; In fact, it's better for us that you keep the original commits so we can see more easily what's changed between each review. Don't worry, the commits will all be squeezed down to a single one when merging with master, so the PR history can be as messy as you want (Still, be a bit careful :) ) Thanks!

Please consider doing this, otherwise I need to review the entire PR instead of the changed parts.

RossBrunton:
Sorry about the force pushing, I didn't see your message before pushing that last one! I guess I'm used to gitlab where rewriting history is handled much more egonomically. You can click the secret hidden link on the word "force-pushed" to see the diff introduced by that commit, but it's nowhere near as good as gitlab's "change history thing".

Anyway, I rushed it a bit, so there may be small mistakes I need to clean up.

src/map.rs
RossBrunton: Tilesets can contain tiles, which can contain objectgroups which contain objects which can contain templates. So while running inside the closure we need access to the cache (in case we need to process another template's tileset). This would mean that the cache is being mutably borrowed twice, once in the closure, and once as the `&mut self` in cache.

I'm not sure how else we could work around this besides having the closure take the cache as an argument?.

src/tileset.rs
RossBrunton: Whoops! What should I call it? The paths are tracked in multiple places and now it can be a map, template or (I think) a tilesheet.

bjorn: The name `root_path` has appeared elsewhere, and I guess it would be appropriate here as well. However, I don't really agree with the term "root", and would also be fine with just `path` (and would assume this path does not include the file name, actually).

aleokdev: The function should be renamed to something that doesn't specify maps (e.g. parse_xml_in_file) and the path parameter couls just be file_path then.

src/cache.rs
```diff
@@ -11,13 +11,7 @@ pub type ResourcePathBuf = PathBuf;
 
 pub trait ResourceCache {
```
aleokdev, line 12: Templates should be cached in the resource cache. Feel free to add any methods neccesary to the trait (Just keep them consistent with the tileset ones)

src/template.rs
```diff
@@ -0,0 +1,122 @@
...
+pub(crate) struct Template {
+    pub path: PathBuf,
+    pub tileset: Option<Arc<Tileset>>,
+    pub object: Option<ObjectData>,
+}
+
+impl Template {
+    pub(crate) fn parse_and_append_template(
+        path: &Path,
+        templates: &mut Vec<Template>,
```
aleokdev, line 26: This should be in the cache, not here.

src/template.rs
```diff
@@ -0,0 +1,142 @@
...
+    pub object: Option<ObjectData>,
+}
+
+impl Template {
+    pub(crate) fn new_external<CacheTy: ResourceCache>(
+        path: &Path,
+        parse_context: &mut ParseContext<CacheTy>,
+    ) -> Result<usize, TiledError> {
+        // First, check if this template already exists in our list of templates (it will be uniquely identified by its
+        // path).
```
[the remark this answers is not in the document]
aleokdev: This function should return `Result<Template, TiledError>`.

src/template.rs
```diff
@@ -0,0 +1,122 @@
...
+        if let Some((id, _)) = templates
+            .iter()
+            .enumerate()
+            .find(|(_, template)| template.path == path)
+        {
+            return Ok(id);
+        }
```
aleokdev, line 37: This can be moved to the caller using `cache.get_or_try_insert_template` (or `get_template` and then `insert_template`)

src/template.rs
```diff
@@ -0,0 +1,122 @@
...
+                    ))
+                }
+                _ => {}
+            }
+        }
+    }
+
+    fn parse_and_append_external_template(
+        parser: &mut impl Iterator<Item = XmlEventResult>,
+        _attrs: &Vec<OwnedAttribute>,
```
aleokdev, line 74: If the attributes aren't being used, they shouldn't be passed to the function.

src/template.rs
```diff
@@ -0,0 +1,122 @@
...
+        parse_tag!(parser, "template", {
+            "object" => |attrs| {
+                object = Some(ObjectData::new(parser, attrs, Some(&tilesets), templates, Some(template_id), template_path, cache)?);
+                Ok(())
+            },
+            "tileset" => |attrs| {
+                let res = Tileset::parse_xml_in_map(parser, attrs, template_path, templates, Some(template_id), cache)?;
+                match res.result_type {
+                    EmbeddedParseResultType::ExternalReference { tileset_path } => {
+                        let tileset = if let Some(ts) = cache.get_tileset(&tileset_path) {
```
aleokdev, line 99: You can use `Option::unwrap_or_else` here... Or `cache.get_or_try_insert_tileset`.

aleokdev (changes requested):

RossBrunton:
So it seems there's a design issue here. Tiles must be able to identify which template they are "from" (so they can do the automatic gid resolving). I implemented this as an index into a vector that is collected as we parse, but that seems to be not the design you are looking for (which makes sense, I don't really like it either). Also complicating things is the fact that tiles inside templates can themselves include templates (Tiles can contain ObjectGroups).

Any suggestions on how I can refactor that so it's less messy?

src/template.rs
```diff
@@ -0,0 +1,122 @@
...
+
+/// A template, consisting of an object and a tileset
+///
+/// The path is the path to the template file; it is used to identify them uniquely, and skip parsing the file multiple
+/// times.
+#[derive(Clone, Debug, PartialEq)]
+pub(crate) struct Template {
+    pub path: PathBuf,
+    pub tileset: Option<Arc<Tileset>>,
+    pub object: Option<ObjectData>,
```
aleokdev, line 20: The `object` attribute is mandatory, it just wasn't specified in the docs (oops): https://github.com/mapeditor/tiled/commit/8f1162d9cd3e0660d14ac83193c795e825a5a78a

aleokdev:
> So it seems there's a design issue here. Tiles must be able to identify which template they are "from" (so they can do the automatic gid resolving). I implemented this as an index into a vector that is collected as we parse, but that seems to be not the design you are looking for (which makes sense, I don't really like it either). Also complicating things is the fact that tiles inside templates can themselves include templates (Tiles can contain ObjectGroups).
> 
> Any suggestions on how I can refactor that so it's less messy?

Yes; If you instead put templates in the cache, then the templates vector won't be needed and as such the template constructor won't need to insert itself anywhere (as the tiles it creates can do that themselves). This would imply making templates public, but I don't see how that is an issue, seeing how sharing templates between maps is a pretty common occurrence.

RossBrunton:
What would LayerTileData::tileset_index be in that case?

aleokdev:
> What would LayerTileData::tileset_index be in that case?

It can be the path of the template, as tiles in templates are uncommon enough to not have any impact on the memory and performance footprint.

aleokdev:
That'd also remove the dependency of the map when accessing the tileset of a template's tile. However, maybe for consistency's sake a Vec of Arc to the map templates can be stored in the map, just like tilesets, and then we could have tiles store the template index in the map. Without looking at the code (I'm on phone rn) I believe that is what your current impl is doing

aleokdev:
From the two options the second one sounds better, as you'd have access to the templates a map is referencing to and overall it'd be more consistent.

RossBrunton:
If I'm understanding you correctly, as part of parsing the template we generate the tile object with an index into the map's templates vector for its tileset.

How do we get that index if the Template constructor doesn't create a new entry in said vector?

bjorn:
Should we potentially store the tile reference for templates differently? We could make the tile reference in tile objects an enum, which can either store a tileset index + tile id, or directly the `Arc<Tileset>` + tile id. The latter would be used for the actual templates, whereas regular tile objects, but probably also tile object template instances, if we "apply" them during load (adding the tileset to the map if it's not already there), could keep using the tileset index + tile id.

aleokdev:
> Should we potentially store the tile reference for templates differently? We could make the tile reference in tile objects an enum, which can either store a tileset index + tile id, or directly the `Arc<Tileset>` + tile id. The latter would be used for the actual templates, whereas regular tile objects, but probably also tile object template instances, if we "apply" them during load (adding the tileset to the map if it's not already there), could keep using the tileset index + tile id.

I like this idea, but should we have an arc to the tileset + id or arc to the template + id? The first is more concrete and has less indirections but the latter has the template available, in case the user needs it.

bjorn:
> I like this idea, but should we have an arc to the tileset + id or arc to the template + id? The first is more concrete and has less indirections but the latter has the template available, in case the user needs it.

I don't really understand the choice, since I think the template needs an `Arc<Tileset>` either way. Whether we need an arc to the template depends on whether we'll copy its properties down to the instance I guess.

aleokdev:
We can remove the copy, this PR is already breaking anyways

RossBrunton:
Sorry for the delay, have made that change now. LayerTileData now has a destructor which may or may not be a problem for performance (tbh, this was why I was reluctant to do it like this initially). When I get a free moment some time in the future I'll merge in the changes from master and respond to other comments.

src/layers/tile/mod.rs
```diff
@@ -15,13 +15,22 @@ mod util;
...
 pub use infinite::*;
 
+/// The location of the tileset this tile is in
+///
+/// Tilesets can be located in either one of the map's tilesets, or a tileset specified by a template.
+#[derive(Clone, Debug, PartialEq)]
+pub(crate) enum TilesetLocation {
+    /// Index into the Map's tileset list, guaranteed to be a valid index of the map tileset container
+    Map(usize),
+    /// Arc of the tileset itself if and only if this is location is from a template
```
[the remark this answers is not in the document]
aleokdev: BTW, there's a typo, there's an extra `is` before `location` :)

src/layers/tile/mod.rs
```diff
@@ -115,18 +136,27 @@ impl<'map> LayerTile<'map> {
...
     /// Use [`LayerTile::get_tile`] if you want to obtain the [`Tile`] that this layer tile is
     /// referencing.
-    pub fn tileset_index(&self) -> usize {
-        self.data.tileset_index
+    pub fn tileset_index(&self) -> Option<usize> {
+        match self.data.tileset_location {
+            TilesetLocation::Map(n) => Some(n),
+            TilesetLocation::Template(_) => None,
+        }
     }
```
aleokdev, line 160: Now that `tileset_index` has been replaced by `tileset_location`, it makes more sense for this function to be replaced by a `tileset_location` readonly property.

src/objects.rs
```diff
@@ -42,26 +43,63 @@ impl ObjectData {
...
-            required: [
-                ("x", x, |v:String| v.parse().ok()),
-                ("y", y, |v:String| v.parse().ok()),
-            ],
+            required: [],
             TiledError::MalformedAttributes("objects must have an x and a y number".to_string())
         );
+
+        // If the template attribute is there, we need to go fetch the template file
+        let template = if let Some(template) = template {
```
aleokdev, line 74: This `if let` can be replaced by a `template.map(...)`.

aleokdev (commented):
Much better! However, I still have some comments, along with some previous remarks that we'll have to iron out before introducing the feature into the public API.

bjorn:
> `LayerTileData` now has a destructor which may or may not be a problem for performance (tbh, this was why I was reluctant to do it like this initially).

Just wondering, but how hard would it be to have this enum only for objects, such that we can at least avoid it (and the destructor) for tile layers, which can never refer to a tileset from a template?

aleokdev:
Maybe it's about time we introduce a `Tile` wrapper that refers to the parent object, instead of just returning a `TileData` from it. This wrapper could handle the enum introduced in the object so that it also has a valid `tileset()` property.

src/layers/tile/infinite.rs
```diff
@@ -92,8 +93,9 @@ impl Chunk {
     pub const TILE_COUNT: usize = Self::WIDTH as usize * Self::HEIGHT as usize;
 
     pub(crate) fn new() -> Self {
+        const INIT: Option<LayerTileData> = None;
```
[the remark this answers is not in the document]
RossBrunton: You can't initialize an array of a non-Copy type using the `[V; C]` syntax... But for some reason you can do this and it works? I'm not sure the specifics.

src/layers/tile/mod.rs
```diff
@@ -40,8 +49,11 @@ impl LayerTileData {
     /// Use [`LayerTile::get_tile`] if you want to obtain the [`Tile`] that this layer tile is
     /// referencing.
     #[inline]
-    pub fn tileset_index(&self) -> usize {
-        self.tileset_index
+    pub fn tileset_index(&self) -> Option<usize> {
```
aleokdev, line 52: This should be `tileset_location` now that the index has been replaced by location, and it should return `TilesetLocation` (should just be a readonly property)

aleokdev (approved):
Really sorry for being so late on reviewing this!
The changes look great! Very happy about how it turned out. There's just a single quirk I've seen, apart from that it looks ready to merge.

aleokdev:
By the way, I completely forgot to switch the base from `master` to `0.11`, which is the version this PR is targeting. Sorry!!! I know there are merge conflicts :(

aleokdev:
Working on the merge.

Here's the bigger picture of the current structure of the PR:
tiledrs drawio(1)

aleokdev:
@RossBrunton I've merged 0.11 with the template branch locally. Could you allow commits from us to push the changes into the PR?

RossBrunton:
There we go, that should be it. Thanks for doing the merge.

aleokdev:
> Just wondering, but how hard would it be to have this enum only for objects, such that we can at least avoid it (and the destructor) for tile layers, which can never refer to a tileset from a template?

Thinking more about this lately. It's a good idea. I'll try to remove `LayerTileData` from `Object`, mostly because it doesn't make sense and, as you said, it does make things much more complicated.

aleokdev:
AFAIK, there is no way to have an object from a tile refer to a tileset from *outside* its environment, so that should simplify things.

E.g. If there is a template which has a tileset which has a tile which has an object which refers to a tileset, the tileset it can refer to can only be the tileset from the object's template... Right?
Screenshot_20220419_163423

There seems to be ambiguity here. I need more information about how object GIDs within tiles work; The documentation is lacking in this regard.

aleokdev:
image

Here's another edge case... Not sure how to deal with this one either. An object can refer to a template that contains itself.

bjorn:
> Disallowing tiles from having tile objects from within them would reduce complexity by a lot. Currently that's the blocker here.

I think we clarified this on Discord. In Tiled, the `ObjectGroup` of a `Tile` is not allowed to refer to tile objects, nor to templates of tile objects. So, these kind of circular references are impossible and don't need to be supported. Probably something that should also be mentioned in the Tiled manual.

aleokdev:
That should be it. Ready for final review.

aleokdev:
I'm more worried about the external interface and assumptions taken, the actual implementation can be upgraded later on without issues.

src/cache.rs
```diff
@@ -23,49 +23,45 @@ pub trait ResourceCache {
...
+    /// # use std::sync::Arc;
+    ///
     /// # fn main() -> Result<()> {
     /// let mut loader = Loader::new();
     /// let path = "assets/tilesheet.tsx";
     ///
     /// assert!(loader.cache().get_tileset(path).is_none());
-    /// loader.load_tmx_map("assets/tiled_base64_external.tmx");
+    /// let tileset = Arc::new(loader.load_tsx_tileset(path)?);
+    /// loader.cache_mut().insert_tileset(path, tileset);
```
aleokdev, line 36: Maybe these changes should be done in a different PR.

src/tileset.rs
```diff
@@ -251,7 +242,12 @@ impl Tileset {
                 Ok(())
             },
             "tile" => |attrs| {
-                let (id, tile) = TileData::new(parser, attrs, &prop.root_path, for_tileset.clone(), reader,cache)?;
+                // Tilesets can have tiles which can have object groups which can have objects that
+                // refer to this tileset. The issue is that this tileset isn't even constructed yet.
+
+                // This is a corner case that isn't dealt with currently.
+                // FIXME
```
bjorn, line 249: Adding outdated comment?

aleokdev (approved):

src/tileset.rs
aleokdev: oops

CHANGELOG.md
```diff
@@ -6,8 +6,14 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## [0.11.0]
+### Added
+- Template support!
+Templates are loaded automatically when they are encountered, and are treated as intermediate
+objects. As such, `ResourceCache` has now methods for both getting and inserting them in. (#170)
```
bjorn:
```suggestion
objects. As such, `ResourceCache` has now methods for both getting and inserting them (#170).
```

src/tileset.rs
```diff
@@ -230,7 +242,12 @@ impl Tileset {
                 Ok(())
             },
             "tile" => |attrs| {
-                let (id, tile) = TileData::new(parser, attrs, &prop.root_path)?;
+                // Tilesets can have tiles which can have object groups which can have objects that
+                // refer to this tileset. The issue is that this tileset isn't even constructed yet.
+
+                // This is a corner case that isn't dealt with currently.
+                // FIXME
+                let (id, tile) = TileData::new(parser, attrs, &prop.root_path,  reader,cache)?;
```
bjorn:
```suggestion
                let (id, tile) = TileData::new(parser, attrs, &prop.root_path, reader, cache)?;
```

src/template.rs
```diff
@@ -0,0 +1,107 @@
...
+                Ok(())
+            },
+            "tileset" => |attrs: Vec<OwnedAttribute>| {
+                let res = Tileset::parse_xml_in_map(parser, &attrs, template_path, reader, cache)?;
+                match res.result_type {
+                    EmbeddedParseResultType::ExternalReference { tileset_path } => {
+                        tileset = Some(if let Some(ts) = cache.get_tileset(&tileset_path) {
+                            ts
+                        } else {
+                            let tileset = Arc::new(crate::parse::xml::parse_tileset( &tileset_path,  reader,cache)?);
```
bjorn:
```suggestion
                            let tileset = Arc::new(crate::parse::xml::parse_tileset(&tileset_path, reader, cache)?);
```

src/tile.rs
```diff
@@ -91,7 +93,9 @@ impl TileData {
                 Ok(())
             },
             "objectgroup" => |attrs| {
-                objectgroup = Some(ObjectLayerData::new(parser, attrs, None)?.0);
+                // Tile objects are not allowed within tile object groups, so we can pass None as the
+                // tilesets vector
+                objectgroup = Some(ObjectLayerData::new(parser, attrs, None, None, path_relative_to, reader,cache)?.0);
```
bjorn:
```suggestion
                objectgroup = Some(ObjectLayerData::new(parser, attrs, None, None, path_relative_to, reader, cache)?.0);
```

src/layers/group.rs
```diff
@@ -22,6 +20,9 @@ impl GroupLayerData {
         infinite: bool,
         map_path: &Path,
         tilesets: &[MapTilesetGid],
+        for_tileset: Option<Arc<Tileset>>,
+        reader: &mut impl ResourceReader,
+        cache: &mut impl ResourceCache,
```
bjorn, line 25: I see that in many places we're passing around both a reader and a cache. Would it make sense to pass around a loader instead?

src/map.rs
```diff
@@ -153,11 +153,18 @@ impl Map {
...
                 match res.result_type {
                     EmbeddedParseResultType::ExternalReference { tileset_path } => {
-                        let tileset = cache.get_or_try_insert_tileset_with(tileset_path.clone(), || crate::parse::xml::parse_tileset(&tileset_path, reader))?;
+                        let tileset = if let Some(ts) = cache.get_tileset(&tileset_path) {
+                            ts
+                        } else {
+                            let tileset = Arc::new(crate::parse::xml::parse_tileset(&tileset_path,  reader, cache)?);
+                            cache.insert_tileset(tileset_path.clone(), tileset.clone());
+                            tileset
+                        };
```
bjorn, line 166: Could we, instead of this construct, have passed the `cache` as an argument to the lambda used by `get_or_try_insert_tileset_with`? Or is that not preferable?

bjorn (commented):
Added a few minor suggestions and some questions.

src/map.rs
aleokdev: Can't do because of borrow rules, get_or_try_insert_with uses &mut cache and passing &cache into the closure wouldn't be possible.

bjorn: I find this quite unexpected. Surely a mutable reference can be turned into a non-mutable reference? Note that I'm asking about passing it in as a parameter, not as a captured value.

aleokdev: Oh I see what you mean now... Yeah, that would be possible. We can try that.

bjorn: Actually never mind, I think it'll still not work, since `parse_tileset` actually needs a mutable reference, and we probably can't create another one with the vacant hash map entry around. And if we work around that, the advantage of using `tilesets.entry` is gone anyway.

src/objects.rs
```diff
@@ -80,33 +185,75 @@ impl ObjectData {
...
+                n.get_or_insert(obj.name.clone());
+                t.get_or_insert(obj.obj_type.clone());
+                if let Some(templ_tile) = obj.tile.clone() {
+                    tile.get_or_insert(templ_tile);
+                }
+                Ok(template)
+            })
+            .transpose()?;
+
+        let x = x.unwrap_or(0f32);
```
aleokdev, line 244: Is this correct? @bjorn

src/layers/group.rs
aleokdev: Not sure, Loader owns the reader/cache

bjorn: How does that pose a problem for passing `loader: &mut Loader` instead of the above two parameters?

src/objects.rs
bjorn: Yes, if `x` or `y` attributes are missing, they default to 0, and for object templates they are never written.

This time [the documentation](https://doc.mapeditor.org/en/stable/reference/tmx-map-format/#object) at least already mentions the default value. :-)

aleokdev (approved):
It still has some internal quirks, but we'll deal with them later. Last review!

src/objects.rs
```diff
@@ -80,33 +185,75 @@ impl ObjectData {
...
+                    let template = Template::parse_template(&template_path, reader, cache)?;
+                    // Insert it into the cache
+                    cache.insert_template(&template_path, template.clone());
+                    template
+                };
+
+                // The template sets the default values for the object
+                let obj = &template.object;
+                x.get_or_insert(obj.x);
+                y.get_or_insert(obj.y);
```
bjorn: Actually a template instance never inherits the `x` or `y` from the template object (these attributes are also never saved for the template object), so this effectively just makes then 0, which is already the default.

src/objects.rs
```diff
@@ -80,33 +185,75 @@ impl ObjectData {
...
+                v.get_or_insert(obj.visible);
+                #[allow(deprecated)]
+                w.get_or_insert(obj.width);
+                #[allow(deprecated)]
+                h.get_or_insert(obj.height);
+                r.get_or_insert(obj.rotation);
+                id.get_or_insert(obj.id);
```
bjorn: The `id` is also never inherited from the template.

src/objects.rs
```diff
@@ -138,6 +285,16 @@ impl ObjectData {
 
         let shape = shape.unwrap_or(ObjectShape::Rect { width, height });
```
bjorn: Actually the shape might inherit from the template object.

bjorn (changes requested):
Sorry, I found a few things could be improved about the actual inheritance from the template.

src/objects.rs
```diff
@@ -283,18 +279,20 @@ impl ObjectData {
             },
         });
 
-        let shape = shape.unwrap_or(ObjectShape::Rect { width, height });
-
         // Possibly copy properties from the template into the object
         // Any that already exist in the object's map don't get copied over
         if let Some(templ) = template {
+            shape.get_or_insert(templ.object.shape.clone());
```
bjorn, line 285: For performance reasons I guess it might make sense to use `get_or_insert_with` for all the cases where `clone()` needs to be called.

bjorn (dismissed):

aleokdev (approved):

aleokdev:
Thank you @RossBrunton and sorry for the wait!

RossBrunton:
No problem, thanks for taking over this and improving it.

PR #174, opened 2022-03-02
Add preliminary documentation

### LINKED ISSUE #151: Better documentation
Add doctests wherever possible and add documentation to every struct that is missing it.

aleokdev:
Adds documentation to all public members of the crate, as well as some doctests. Closes #151.

### DIFF
--- a/README.md
+++ b/README.md
@@ -9,7 +9,7 @@ Code contributions are welcome as are bug reports, documentation, suggestions an
 
 [There is a package on crates.io](https://crates.io/crates/tiled), to use simply add:
 
-```
+```toml
 tiled = "0.9.5"
 ```
 
@@ -20,13 +20,18 @@ The minimum supported TMX version is 0.13.
 ### Example
 
 ```rust
-use tiled::Map;
+use tiled::{FilesystemResourceCache, Map};
 
 fn main() {
-    let map = Map::parse_file("assets/tiled_base64_zlib.tmx").unwrap();
+    let map = Map::parse_file(
+        "assets/tiled_base64_zlib.tmx",
+        &mut FilesystemResourceCache::new(),
+    )
+    .unwrap();
     println!("{:?}", map);
-    println!("{:?}", map.tileset_by_gid(22));
+    println!("{:?}", map.tilesets()[0].get_tile(0).unwrap().probability());
 }
+
 ```
 
 ### Licences
--- a/README.md
+++ b/README.md
@@ -1,19 +1,17 @@
 # rs-tiled
-
-[![Rust](https://github.com/mapeditor/rs-tiled/actions/workflows/rust.yml/badge.svg)](https://github.com/mapeditor/rs-tiled/actions/workflows/rust.yml)
-[![Crates.io](https://img.shields.io/crates/v/tiled.svg)](https://crates.io/crates/tiled)
-
-Read maps from the [Tiled Map Editor](http://www.mapeditor.org/) into rust for use in video games. It is game engine agnostic and pretty barebones at the moment. Documentation is available [on docs.rs](https://docs.rs/tiled/).
-
-Code contributions are welcome as are bug reports, documentation, suggestions and criticism.
-
-[There is a package on crates.io](https://crates.io/crates/tiled), to use simply add:
-
 ```toml
 tiled = "0.9.5"
 ```
 
-to the dependencies section of your Cargo.toml.
+[![Rust](https://github.com/mapeditor/rs-tiled/actions/workflows/rust.yml/badge.svg)](https://github.com/mapeditor/rs-tiled/actions/workflows/rust.yml)
+[![Crates.io](https://img.shields.io/crates/v/tiled.svg)](https://crates.io/crates/tiled)
+
+A crate for reading TMX (map) and TSX (tileset) files from the [Tiled Map Editor](http://www.mapeditor.org/) into Rust.
+It provides a huge set of features as well as a strong wrapper over internal features such as GIDs.
+
+Documentation is available [on docs.rs](https://docs.rs/tiled/).
+
+Code contributions are welcome as are bug reports, documentation, suggestions and criticism.
 
 The minimum supported TMX version is 0.13.
 
--- a/README.md
+++ b/README.md
@@ -36,6 +36,6 @@ fn main() {
 
 ### Licences
 
-assets/tilesheet.png by Buch (https://opengameart.org/content/sci-fi-interior-tiles)
+assets/tilesheet.png by [Buch](https://opengameart.org/content/sci-fi-interior-tiles)
 
 Licenced under MIT
--- a/src/image.rs
+++ b/src/image.rs
@@ -15,8 +15,11 @@ pub struct Image {
     /// [source]: https://doc.mapeditor.org/en/stable/reference/tmx-map-format/#image
     // TODO: Embedded images
     pub source: PathBuf,
+    /// The width in pixels of the image.
     pub width: i32,
+    /// The height in pixels of the image.
     pub height: i32,
+    /// A color that should be interpreted as transparent (0 alpha), if any.
     pub transparent_colour: Option<Color>,
 }
 
--- a/src/image.rs
+++ b/src/image.rs
@@ -4,6 +4,7 @@ use xml::attribute::OwnedAttribute;
 
 use crate::{error::TiledError, properties::Color, util::*};
 
+/// A reference to an image stored somewhere within the filesystem.
 #[derive(Debug, PartialEq, Eq, Clone)]
 pub struct Image {
     /// The filepath of the image.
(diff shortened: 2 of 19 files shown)


src/layers/group.rs
```diff
@@ -81,9 +81,12 @@ impl GroupLayerData {
 map_wrapper!(GroupLayer => GroupLayerData);
 
 impl<'map> GroupLayer<'map> {
+    /// Returns an iterator over the layers present in this group, ordered in the same way as in the
+    /// TMX file it was loaded from.
```
bjorn, line 85: I'm wondering if it's really helpful to document that the order matches that of the file. It should be expected behavior. Maybe at least remove the "TMX" since otherwise we have to remember to remove it when adding support for the JSON format. We also need to remember to reword it when this library allows dynamic creation of maps.

Alternatively, we'd document the order matches the display order of the layers, so back to front.

src/layers/tile/infinite.rs
```diff
@@ -82,13 +82,19 @@ fn tile_to_chunk_pos(x: i32, y: i32) -> (i32, i32) {
...
 
 #[derive(Debug, PartialEq, Clone)]
-pub struct Chunk {
+pub(crate) struct Chunk {
     tiles: Box<[Option<LayerTileData>; Self::TILE_COUNT]>,
 }
 
 impl Chunk {
+    /// Internal infinite layer chunk width. Do not rely on this value as it might change between
+    /// versions.
```
bjorn, line 91: With the `Chunk` struct changed to `pub(crate)`, is this warning to not rely on the value staying the same still relevant?

src/lib.rs
```diff
@@ -11,6 +14,7 @@ mod tileset;
 mod util;
 
 pub use animation::*;
+
```
bjorn, line 17: Stray empty line added?

src/objects.rs
```diff
@@ -19,6 +23,8 @@ pub enum ObjectShape {
 }
 
 /// Raw data belonging to an object. Used internally and for tile collisions.
+/// 
+/// Also see the [TMX docs](https://doc.mapeditor.org/en/stable/reference/tmx-map-format/#tmx-object).
```
bjorn, line 27: We don't need `#[allow(missing_docs)]` here, even though the members are not documented?

src/util.rs
```diff
@@ -54,8 +54,9 @@ macro_rules! parse_tag {
 
 /// Creates a new type that wraps an internal data type over along with a map.
 macro_rules! map_wrapper {
-    ($name:ident => $data_ty:ty) => {
+    ($(#[$attrs:meta])* $name:ident => $data_ty:ty) => {
         #[derive(Clone, PartialEq, Debug)]
+        $(#[$attrs])*
```
bjorn, line 59: That is some magic that I won't try to understand right now... :-)

README.md
```diff
@@ -20,13 +20,18 @@ The minimum supported TMX version is 0.13.
...
 
 ```rust
-use tiled::Map;
+use tiled::{FilesystemResourceCache, Map};
 
 fn main() {
-    let map = Map::parse_file("assets/tiled_base64_zlib.tmx").unwrap();
+    let map = Map::parse_file(
+        "assets/tiled_base64_zlib.tmx",
+        &mut FilesystemResourceCache::new(),
```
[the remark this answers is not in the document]
aleokdev: Yes, it is a candidate for lifetime extension.

src/image.rs
```diff
@@ -15,8 +15,11 @@ pub struct Image {
     /// [source]: https://doc.mapeditor.org/en/stable/reference/tmx-map-format/#image
     // TODO: Embedded images
     pub source: PathBuf,
+    /// The width in pixels of the image.
     pub width: i32,
+    /// The height in pixels of the image.
     pub height: i32,
+    /// A color that should be interpreted as transparent (0 alpha), if any.
     pub transparent_colour: Option<Color>,
```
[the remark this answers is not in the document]
aleokdev: Yes, opened an issue about it. Changing it in another PR though.

src/layers/tile/infinite.rs
aleokdev: Ah, shouldn't have changed the visibility here, my bad.

src/lib.rs
aleokdev: Oops.

src/properties.rs
```diff
@@ -55,16 +57,26 @@ impl FromStr for Color {
...
 #[derive(Debug, PartialEq, Clone)]
 pub enum PropertyValue {
+    /// A boolean value. Corresponds to the `bool` property type.
     BoolValue(bool),
+    /// A floating point value. Corresponds to the `float` property type.
     FloatValue(f32),
+    /// A signed integer value. Corresponds to the `int` property type.
     IntValue(i32),
+    /// A color value. Corresponds to the `color` property type.
     ColorValue(u32),
```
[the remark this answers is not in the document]
aleokdev: Another issue to fix. Never got to changing it.

src/util.rs
aleokdev: It's not very complex as long as you know about token repetition with `*`; First we match against any attributes, which are meta elements encased by `#[ ... ]`, and then we repackage them herebin order to serve them as attributes for the wrapper.

README.md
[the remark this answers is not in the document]
aleokdev: More specifically, the lifetime of the borrow is unspecified and so the compiler fills it in from the function declaration, which asserts that it must last at least as long as the function call by default.

src/objects.rs
aleokdev: I added it because the members are all pretty self-explanatory from the TMX docs linked. (It's there because of the members, not the parent enum).

bjorn: I don't mind the link to the TMX docs. My comment was not really referring to that line, but rather wondering why we didn't need `#[allow(missing_docs)]` for this struct, even though it doesn't have docs for its members.

aleokdev: But that attribute is present on the struct. What do you mean?

bjorn: This change adds that attribyte to the `ObjectShape` enum, but I do not see it on the `ObjectData` struct. Am I blind?

aleokdev: Oh lol, now I see what you mean. I thought this comment was referring to ObjectShape. No, docs are missing, that's why this PR is a draft. I still haven't finished some parts.

src/layers/group.rs
```diff
@@ -78,11 +78,15 @@ impl GroupLayerData {
     }
 }
 
-map_wrapper!(GroupLayer => GroupLayerData);
+map_wrapper!(
+    #[doc = "A group layer, used to organize the layers of the map in a hierarchy."]
+    #[doc = "## Note"]
+    #[doc = "This type is currently nonstandard, as its attributes do not recursively affect child layers."]
```
bjorn, line 84: I think they shouldn't, and that this is up to the library or application implementing the rendering.

Also, suppose you'd render a Tiled map by replicating it in an entity hierarchy. Then, child entities (layers or objects) could naturally inherit transformations and visibility from their parent entities. Doing it in this library as well in such a case would only make things more complicated.

In Tiled, I do have a few helper functions that are used where needed, like `Layer::effectiveOpacity`, `Layer::effectiveParallaxFactor`, etc.

aleokdev: Then I think the documentation should be updated as well. Right now it reads as if any implementation must ensure that the children also have the properties of the parent: https://doc.mapeditor.org/en/stable/reference/tmx-map-format/#group
Specifying "in the editor" would be helpful.

bjorn: Hmm, maybe, though that makes it sound strangely specific to the editor, as if it would not be the expected behavior elsewhere. The documentation is trying to communicate the expected behavior, but I don't think it needs to specify whether that is implemented in the loader, the renderer or the application. It's our choice what to do in `rs-tiled`, and I think we should leave that to the user of the library.

src/layers/group.rs
```diff
@@ -80,8 +80,10 @@ impl GroupLayerData {
 
 map_wrapper!(
     #[doc = "A group layer, used to organize the layers of the map in a hierarchy."]
+    #[doc = "\nAlso see the [TMX docs](https://doc.mapeditor.org/en/stable/reference/tmx-map-format/#group)."]
     #[doc = "## Note"]
-    #[doc = "This type is currently nonstandard, as its attributes do not recursively affect child layers."]
+    #[doc = "Although the TMX documentation states that its attributes recursively affect child
+    layers, this is currently only true for the editor, and not the crate."]
```
bjorn, line 86: Maybe something like:

"In Tiled, the properties of the group layer recursively affect child layers. Implementing this behavior is left up to the user of this library."

bjorn (approved):

PR #182, opened 2022-03-04
Implement Deref for wrapper types

### LINKED ISSUE #168: More property/member consistency
Right now, there are some members that shouldn't be public, as they should be protected against mutation; e.g. `Map::width/height/tile_width`. These should be changed to properties.

aleokdev:
An experiment. Exposes data types but implements `Deref<Target=TData>` for wrapper types, leaving practically the same interface but also allowing the user to also use data types in more performance-critical parts.

Closes #168.


src/objects.rs
```diff
@@ -28,92 +28,42 @@ pub enum ObjectShape {
...
-        self.name.as_ref()
-    }
-
-    /// The type of the object, which is arbitrary and set by the user.
-    #[inline]
-    pub fn obj_type(&self) -> &str {
-        self.obj_type.as_ref()
-    }
-
+    pub name: String,    /// The type of the object, which is arbitrary and set by the user.
```
bjorn, line 34: Missing newline before comment.

src/layers/tile/infinite.rs
```diff
@@ -9,6 +9,7 @@ use crate::{
 
 use super::util::parse_data_line;
 
+/// The raw data of a [`InfiniteTileLayer`]. Does not include a reference to its parent [`Map`](crate::Map).
```
bjorn, line 12: Does it make sense to say something about the parent here? It sounds more suitable as documentation for `TileLayerData` to me.

src/layers/tile/finite.rs
```diff
@@ -7,6 +7,7 @@ use crate::{
 
 use super::util::parse_data_line;
 
+/// The raw data of a [`FiniteTileLayer`]. Does not include a reference to its parent [`Map`](crate::Map).
```
bjorn, line 10: Same here, the section about the parent is probably only relevant for `TileLayerData`.

bjorn (commented):
Just some comments.

bjorn:
And just to add, my first impression of this change is very good. I love that we can get rid of all those getters again!

src/layers/tile/finite.rs
aleokdev: It's relevant to all map wrapped objects; they all have the same docs but with different type references

bjorn: Ah, right.

aleokdev:
Note: Getters have stayed for members whose state has some kind of runtime guarantee (e.g. Object ID must be local within map) or  rely on the state of another object (e.g. tile layer width/height depend on the size of the tiles vector and vice-versa)

aleokdev:
CI is broken again. Same error as when I added the SFML example, it's using some old version of the crate instead of the PR one for some reason. Sigh.

aleokdev:
It doesn't even try to build it for some reason...

aleokdev:
Apparently it needed the branch to be merged with master.

src/util.rs
```diff
@@ -74,6 +74,14 @@ macro_rules! map_wrapper {
                 self.map
             }
         }
+
+        impl<'map> std::ops::Deref for $name<'map> {
+            type Target = $data_ty;
+
+            fn deref(&self) -> &'map Self::Target {
```
bjorn, line 81: Should this thing be `#[inline]`?

src/layers/object.rs
```diff
@@ -12,7 +12,8 @@ use crate::{
 #[derive(Debug, PartialEq, Clone)]
 pub struct ObjectLayerData {
     /// The objects present in this layer.
-    pub objects: Vec<ObjectData>,
+    // TODO: read-only property
```
bjorn, line 15: So this TODO is for a follow-up change?

src/objects.rs
```diff
@@ -27,10 +27,8 @@ pub enum ObjectShape {
 /// Also see the [TMX docs](https://doc.mapeditor.org/en/stable/reference/tmx-map-format/#tmx-object).
 #[derive(Debug, PartialEq, Clone)]
 pub struct ObjectData {
-    /// ID of the object, which is unique per map since Tiled 0.11.
-    ///
-    /// On older versions this value is defaulted to 0.
-    pub id: u32,
+    id: u32,
+    // TODO: make public, differ somewhat from get_tile
```
bjorn, line 31: Why not make it public in this change? Is it more involved?

bjorn (approved):
Approved, but had some comments.

src/util.rs
aleokdev: Oops yeah, definitely

src/objects.rs
aleokdev: Whether it is public or not depends on if we should check for ID uniqueness and fail if there are duplicates.

bjorn: I think it's fine to have `id` private, but this comment is about `tile`.

aleokdev: Oh. Yeah, forgot to make it public. The issue is the name, should rename it to tile_data or something.

src/layers/tile/mod.rs
```diff
@@ -176,7 +163,7 @@ impl<'map> TileLayer<'map> {
     }
 
     /// Obtains the tile present at the position given.
-    /// 
+    ///
```
aleokdev, line 166: Unrelated change

src/layers/tile/finite.rs
```diff
@@ -67,23 +80,11 @@ map_wrapper!(
 
 impl<'map> FiniteTileLayer<'map> {
     /// Obtains the tile present at the position given.
-    /// 
+    ///
```
aleokdev, line 83: Unrelated change

src/layers/tile/mod.rs
```diff
@@ -17,19 +17,39 @@ pub use infinite::*;
...
+    /// Use [`LayerTile::get_tile`] if you want to obtain the [`Tile`] that this layer tile is
+    /// referencing.
+    #[inline]
+    pub fn tileset_index(&self) -> usize {
+        self.tileset_index
+    }
+
+    /// Get the layer tile's local id within its parent tileset.
+    #[inline]
+    pub fn id(&self) -> u32 {
```
aleokdev, line 49: Add guarantees

src/layers/object.rs
aleokdev: No, will do later, forgot to impl it

src/layers/object.rs
```diff
@@ -45,6 +44,13 @@ impl ObjectLayerData {
         });
         Ok((ObjectLayerData { objects, colour: c }, properties))
     }
+
+    /// Returns the data belonging to the objects contained within the layer, in no particular
+    /// order.
```
bjorn, line 49: Why would this be "in no particular order", whereas `objects()` will iterate "in the order they were declared in the TMX file"? The order is the same, right?

The order in the TMX file is relevant, btw. It is the rendering order, for objects layers with `draworder` set to `index`, and it can then be manually controlled in Tiled. But I'm not sure how much of this needs to be documented here.

bjorn (approved):
Approved, but might want to tweak the docs for `objects`.

bjorn:
Pretty amazing that the build check even detects errors in commented example code.

aleokdev:
> Pretty amazing that the build check even detects errors in commented example code.

That's because it's not a comment, it's a doctest, and it gets compiled (and executed) when `cargo test` is run.

bjorn (approved):

bjorn:
> That's because it's not a comment, it's a doctest, and it gets compiled (and executed) when `cargo test` is run.

It looks like some code example in a comment though. How does it become a doctest? Are all code blocks in documentation comments doctests?

aleokdev:
> It looks like some code example in a comment though. How does it become a doctest? Are all code blocks in documentation comments doctests?

Lines starting with `///` (or blocks starting with `/**` aren't comments; They are actually converted to `#[doc = "..."]` attributes. This is how cargo can detect docs when generating them. Code in doc attributes is compiled and executed exactly like tests. Here's more info: https://doc.rust-lang.org/rustdoc/the-doc-attribute.html

Issue #183, opened 2022-03-04
Follow RFC 356


aleokdev:
Avoid redundant prefixes: Most notably, in `TiledError`. See https://github.com/rust-lang/rfcs/pull/356

aleokdev:
https://rust-lang.github.io/api-guidelines/interoperability.html

aleokdev:
[Rust API Guidelines](https://rust-lang.github.io/api-guidelines/about.html) to follow for the 1.0.0 release:

# Naming
- [ ] C-CASE
- [ ] C-CONV
- [ ] C-GETTER
- [ ] C-ITER
- [ ] C-ITER-TY
- [ ] C-FEATURE
- [ ] C-WORD-ORDER

# Interoperability
- [ ] C-COMMON-TRAITS
- [ ] C-CONV-TRAITS
- [ ] C-COLLECT
- [ ] C-SERDE
- [ ] C-SEND-SYNC
- [ ] C-GOOD-ERR
- [ ] C-NUM-FMT
- [ ] C-RW-VALUE

# Macros
None in public interface.

# Documentation
- [ ] C-CRATE-DOC
- [ ] C-EXAMPLE
- [ ] C-QUESTION-MARK
- [ ] C-FAILURE
- [ ] C-LINK
- [ ] C-METADATA
- [ ] C-RELNOTES
- [ ] C-HIDDEN

# Predictability
- [ ] C-SMART-PTR
- [ ] C-CONV-SPECIFIC
- [ ] C-METHOD
- [ ] C-NO-OUT
- [ ] C-OVERLOAD
- [ ] C-DEREF
- [ ] C-CTOR

# Flexibility
- [ ] C-INTERMEDIATE
- [ ] C-CALLER-CONTROL
- [ ] C-GENERIC
- [ ] C-OBJECT

# Type safety
- [ ] C-NEWTYPE
- [ ] C-CUSTOM-TYPE
- [ ] C-BITFLAG
- [ ] C-BUILDER

# Dependability
- [ ] C-VALIDATE
- [ ] C-DTOR-FAIL
- [ ] C-DTOR-BLOCK

# Debuggability
- [ ] C-DEBUG
- [ ] C-DEBUG-NONEMPTY

# Future proofing
- [ ] C-SEALED
- [ ] C-STRUCT-PRIVATE
- [ ] C-NEWTYPE-HIDE
- [ ] C-STRUCT-BOUNDS

# Necessities
- [ ] C-STABLE
[...]

PR #185, opened 2022-03-05
Rename `TiledError` to `Error` + Add `Result` type

aleokdev:
Renames `TiledError` to `Error`, following https://github.com/rust-lang/rfcs/pull/356. Also adds a result type.

Part of #183.


src/error.rs
```diff
@@ -51,56 +51,59 @@ pub enum TiledError {
     },
 }
 
-impl fmt::Display for TiledError {
-    fn fmt(&self, fmt: &mut fmt::Formatter<'_>) -> Result<(), fmt::Error> {
+/// A result with an error variant of [`TiledError`].
```
bjorn, line 54:
```suggestion
/// A result with an error variant of [`tiled::error::Error`].
```

bjorn (approved):
Nice cleanup!

bjorn (approved):

PR #190, opened 2022-03-08
Update version to 0.10.0

aleokdev:
(Hopefully) Last PR for this version!

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,7 +1,6 @@
 [package]
-
 name = "tiled"
-version = "0.9.5"
+version = "0.10.0"
 description = "A rust crate for loading maps created by the Tiled editor"
 categories = ["game-development"]
 keywords = ["gamedev", "tiled", "tmx", "map"]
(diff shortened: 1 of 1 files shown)


Cargo.toml
```diff
@@ -18,6 +18,7 @@ default = ["zstd"]
 name = "tiled"
 path = "src/lib.rs"
 
+
```
bjorn, line 21: Maybe not add a newline here?

Cargo.toml
```diff
@@ -10,6 +9,7 @@ readme = "README.md"
 license = "MIT"
 authors = ["Matthew Hall <matthew@quickbeam.me.uk>"]
 edition = "2018"
+include = ["src/**/*.rs", "README.md", "LICENSE"]
```
bjorn, line 12: Might it make sense to include the `CHANGELOG.md`?

bjorn (approved):

Cargo.toml
aleokdev: Maybe, not sure, since these are files that are going to be uploaded to the package. Normally an user that'd like to see all the new changes would check out the repo, ig.

bjorn (approved):

PR #193, opened 2022-03-10
Add loader type

### LINKED ISSUE #187: Create constructors with a defaulted cache
Right now all of our examples and test code uses a default `&mut FilesystemResourceCache` for loading, as it only handles a single map. It would be wise to expose a loading method that doesn't take in a cache and another one that does (with a `_cached` postfix) as it would confuse new users less and would result in less boilerplate for small applications and whatnot.

aleokdev:
Closes #187, and improves the API a lot by doing so.


bjorn (commented):
Wow, looks great to me! I guess just `README.md` and `CHANGELOG.md` updates missing.

tests/lib.rs
```diff
@@ -419,3 +419,7 @@ fn test_group_layers() {
         layer_tile_3.properties.get("key")
     );
 }
+
+mod libtest {
```
bjorn, line 423: What's that for?

bjorn (approved):

tests/lib.rs
aleokdev: Oops. When making doctests I test them out out of the comment so it's easier to work with them. This is a byproduct, forgot to remove it.

bjorn (approved):

PR #197, opened 2022-03-20
Common layer interface

### LINKED ISSUE #179: Add common interface for finite/infinite tile layers
Well, uh. I have no idea how I missed this during the tile layers refactor, but apparently whether a map is infinite or not is defined in the map itself, not the layers, and you can't mix finite layers with infinite ones. This is a pretty big deal for the interface, because it means we could theorically simplify all the finite/infinite stuff into a single variant in `Map` (instead of one variant per layer).

Another idea (bjorn's) would be to just remove the branching entirely by presenting the user with the same interface for both finite and infinite layers. This would have a reaaaally minor overhead but the result would definitely be worth it.

Maybe it's worth a refactor, since the current tile layer interface isn't very good. Putting this issue here just to remind myself that this is a thing.

aleokdev:
Closes #179.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,11 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [0.10.2]
+### Added
+- `TileLayer::width` & `TileLayer::height` for ergonomic access of width/height.
+- `FiniteTileLayerData::get_tile_data`, `InfiniteTileLayerData::get_tile_data`.
+
 ## [0.10.1]
 ### Added
 - `Loader` type for loading map and tileset files without having to necessarily mention the cache
--- a/src/layers/tile/finite.rs
+++ b/src/layers/tile/finite.rs
@@ -62,7 +62,12 @@ impl FiniteTileLayerData {
         })
     }
 
-    pub(crate) fn get_tile(&self, x: i32, y: i32) -> Option<&LayerTileData> {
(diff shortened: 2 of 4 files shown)


bjorn (approved):
The changes look good to me, though I don't quite understand why it closes #179, since the common interface appears to be limited to getting the width/height, which is then non-existent for infinite layers.

In that sense, actually I think a `bounds` property returning a rectangle might be more useful. It would be `rect(0, 0, width, height)` for finite layers, but could return the bounding rectangle of the content for infinite layers, providing a straight-forward way of determining which area to process for tiles while rendering.

#179 also mentions the unified storage for finite and infinite tile layers, but I guess that's an idea better split off into its own issue.

aleokdev:
> The changes look good to me, though I don't quite understand why it closes #179, since the common interface appears to be limited to getting the width/height, which is then non-existent for infinite layers.

Well, that's really all they have to offer. There's already a get_tile method iirc so what was missing were the w/h properties.

> In that sense, actually I think a `bounds` property returning a rectangle might be more useful. It would be `rect(0, 0, width, height)` for finite layers, but could return the bounding rectangle of the content for infinite layers, providing a straight-forward way of determining which area to process for tiles while rendering.

Sure, but this would be an even better job for the future tiles iterator. I'll add a bounds function though, it does seem useful.

> #179 also mentions the unified storage for finite and infinite tile layers, but I guess that's an idea better split off into its own issue.

Yeah I don't think that's happening at any point, since it'd require another refactor which would completely change the interface again. It's ok as is, I guess.

aleokdev:
I'm merging now and introducing the `bounds` function in another PR, since right now neither of the finite/infinite tile layers have the function so there's nothing to unify (it's just a new convenience function). I think it deserves its own PR

PR #200, opened 2022-03-22
Fix iterator lifetimes

aleokdev:
This PR fixes some small lifetime issues with moderately big implications. Currently, code like this fails:
```rs
use tiled::Object;

let spawnpoints: Vec<Object> = map
    .layers()
    .filter_map(|layer| match layer.layer_type() {
        tiled::LayerType::ObjectLayer(layer) => Some(layer),
        _ => None,
    })
    .flat_map(|layer| layer.objects())
    .filter(|object| object.obj_type == "spawn")
    .collect();

dbg!(spawnpoints);
```
The compiler complains with:
```rs
error[E0515]: cannot return reference to function parameter `layer`
  --> src/layers/object.rs:88:23
   |
17 |     .flat_map(|layer| layer.objects())
   |                       ^^^^^^^^^^^^^^^ returns a reference to data owned by the current function

error: aborting due to previous error
```
This is due to the declaration of `impl ExactSizeIterator` functions:
```rs
pub fn objects(&self) -> impl ExactSizeIterator<Item = Object>
```
Which, expanded, look like this:
```rs
pub fn objects<'s>(&'s self) -> impl ExactSizeIterator<Item = Object<'s>> + 's
```
Which limit the lifetime of the iterator and its items to the mapwrapper object. This is not required because the mapwrapper object is just a reference tuple, and so these functions should be like this:
```rs
pub fn objects(&self) -> impl ExactSizeIterator<Item = Object<'map>> + 'map
```
This PR fixes these issues and enforces the fix with some doctests.

### DIFF
--- a/src/layers/group.rs
+++ b/src/layers/group.rs
@@ -90,10 +90,14 @@ map_wrapper!(
 
 impl<'map> GroupLayer<'map> {
     /// Returns an iterator over the layers present in this group in display order.
-    pub fn layers(&self) -> impl ExactSizeIterator<Item = Layer> {
-        self.layers
+    /// ## Example
+    /// TODO
+    pub fn layers(&self) -> impl ExactSizeIterator<Item = Layer<'map>> + 'map {
+        let map: &'map crate::Map = self.map;
+        self.data
+            .layers
             .iter()
-            .map(move |layer| Layer::new(self.map, layer))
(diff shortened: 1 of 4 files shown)


bjorn:
Nice fix and thanks for the great explanation of the issue!

aleokdev:
Why is a commit missing from the timeline?

PR #203, opened 2022-03-26
Remove `Layer` suffix from `LayerType` variants

aleokdev:
Done to improve readability:
```rs
// Previously
match &layer.layer_type() {
    LayerType::TileLayer(l) => do_thing(),
    LayerType::ObjectLayer(l) => do_thing_2(),
    _ => ()
}
// Now
match &layer.layer_type() {
    LayerType::Tile(l) => do_thing(),
    LayerType::Object(l) => do_thing_2(),
    _ => ()
}
```

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -14,8 +14,11 @@ jobs:
     runs-on: ubuntu-latest
 
     steps:
-    - uses: actions/checkout@v2
-    - uses: Swatinem/rust-cache@v1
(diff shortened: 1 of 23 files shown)


bjorn:
Since we're renaming anyway, would it make sense to use `Tiles` and `Objects`, since there layers contain multiples, whereas an `Image` layer does not, and `Group` already indicates multiple?

bjorn (approved):

Issue #167, opened 2022-02-20
Add a `tiles` iterator to `FiniteTileLayer`/`InfiniteTileLayer`


alexdevteam:
Add some kind of iterator that has `Item = (LayerTile, (i32, i32))` or similar. Very useful for rendering code.

bjorn:
But in which order should it iterate the tiles? In Tiled I have:
```c++
    using RenderTileCallback = std::function<void (QPoint, const QPointF &)>;

    /**
     * Calls the given \a renderTile callback for each tile in the given
     * \a exposed rectangle.
     *
     * The callback takes two arguments:
     *
     * \list
     * \li \c tilePos - The tile position of the cell being rendered.
     * \li \c screenPos - The screen position of the cell being rendered.
     * \endlist
     */
    virtual void drawTileLayer(const RenderTileCallback &renderTile,
                               const QRectF &exposed) const = 0;
```
But the order in which it calls `renderTile` depends on the map orientation. Of course this function is doing a bit more than iterating tiles, it also says where each tile goes on the screen.

Note that it doesn't get a tile reference on purpose. There was no tile layer reference going in either. This is to make the function more generally useful, since you could be drawing some other overlay or stacks of tiles based on this.

alexdevteam:
IMO the order shouldn't be specified, and we should leave it as implementation-dependent. This is because we can't really specify the order of iteration of infinite tile layers without adding overhead, for instance.

> Note that it doesn't get a tile reference on purpose. There was no tile layer reference going in either. This is to make the function more generally useful, since you could be drawing some other overlay or stacks of tiles based on this.

I don't really see how it makes it more useful. If you have a tiles iterator the normal thing would be to iterate over the tiles in the layer. Otherwise you could just use two nested fors.

bjorn:
> IMO the order shouldn't be specified, and we should leave it as implementation-dependent. This is because we can't really specify the order of iteration of infinite tile layers without adding overhead, for instance.

It is about the order of the visible tiles. The function takes the exposed rectangle (the area where tiles should get drawn), so the size of the layer is irrelevant and could be infinite. The tile rendering order is important in case they overlap.

> I don't really see how it makes it more useful. If you have a tiles iterator the normal thing would be to iterate over the tiles in the layer. Otherwise you could just use two nested fors.

[...]

PR #210, opened 2022-04-14
Add basic chunk utilities to infinite layers

aleokdev:
As requested by @ShikuWorld & @adambiltcliffe in #167. I'd appreciate if you could review these changes as well, to see if they suit your needs.

Adds map-wrapped chunks (`ChunkWrapper`) and the ability to access infinite tile layer chunks via `InfiniteTileLayerData::chunks` & `InfiniteTileLayerData::chunk_data`, as well as chunk data via `Chunk::get_tile_data` & `ChunkWrapper::get_tile`.

### DIFF
--- a/src/layers/tile/infinite.rs
+++ b/src/layers/tile/infinite.rs
@@ -76,6 +76,18 @@ impl InfiniteTileLayerData {
             })
             .flatten()
     }
+
+    /// Returns an iterator over only the data part of the chunks of this tile layer.
+    ///
+    /// In 99.99% of cases you'll want to use [`Self::chunks()`] instead; Using this method is only
+    /// needed if you *only* require the tile data of the chunks (and no other utilities provided by
+    /// the map-wrapped [`LayerTile`]), and you are in dire need for that extra bit of performance.
+    ///
+    /// This iterator doesn't have any particular order set.
+    #[inline]
+    pub fn chunk_data(&self) -> impl ExactSizeIterator<Item = ((i32, i32), &Chunk)> {
+        self.chunks.iter().map(|(pos, chunk)| (*pos, chunk))
+    }
 }
 
 fn tile_to_chunk_pos(x: i32, y: i32) -> (i32, i32) {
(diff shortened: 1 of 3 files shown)


ShikuWorld (dismissed):
This is perfect, exactly what I needed, thank you!

Not sure what you think of this but at work I like to put an issue number after every TODO to not loose track of them and also to resist my urge to plaster everything with TODO (cause creating an issue is effort).

aleokdev:
True, but since I don't know whether the tiles iterator issue is going to be fixed for 0.11.0, I'll mark the TODOs with the version instead of the issue.

adambiltcliffe (approved):
This looks like it covers my use case just fine, thanks for acting on it so quickly!

aleokdev:
Until this PR is included in the next version, remember you can add
```toml
[patch.crates-io]
tiled = { git = "https://github.com/aleokdev/rs-tiled", branch = "chunks-iter" }
```
To your `Cargo.toml` file to be able to use the changes in the PR.

ShikuWorld:
> Until this PR is included in the next version, remember you can add
> 
> ```toml
> [patch.crates-io]
> tiled = { git = "https://github.com/aleokdev/rs-tiled", branch = "chunks-iter" }
> ```
> 
> To your `Cargo.toml` file to be able to use the changes in the PR.

Thank you I actually didn't know that.

src/layers/tile/infinite.rs
```diff
@@ -167,4 +215,51 @@ impl<'map> InfiniteTileLayer<'map> {
...
+
+    /// Returns an iterator over different parts of this map called [`Chunk`]s.
+    ///
+    /// These **may not** correspond with the chunks in the TMX file, as the chunk size is
+    /// implementation defined (Check [`Chunk::WIDTH`], [`Chunk::HEIGHT`]).
+    ///
+    /// The iterator item contains the position of the chunk in chunk coordinates along with a
+    /// reference to the actual chunk at that position.
+    ///
+    /// This iterator doesn't have any particular order set.
```
bjorn, line 227:
```suggestion
    /// This iterator doesn't have any particular order.
```

src/layers/tile/infinite.rs
```diff
@@ -76,6 +76,18 @@ impl InfiniteTileLayerData {
...
             .flatten()
     }
+
+    /// Returns an iterator over only the data part of the chunks of this tile layer.
+    ///
+    /// In 99.99% of cases you'll want to use [`Self::chunks()`] instead; Using this method is only
+    /// needed if you *only* require the tile data of the chunks (and no other utilities provided by
+    /// the map-wrapped [`LayerTile`]), and you are in dire need for that extra bit of performance.
+    ///
+    /// This iterator doesn't have any particular order set.
```
bjorn, line 86: Maybe remove "set" at the end of the sentence?
```suggestion
    /// This iterator doesn't have any particular order.
```

src/layers/tile/infinite.rs
```diff
@@ -85,28 +97,64 @@ fn tile_to_chunk_pos(x: i32, y: i32) -> (i32, i32) {
...
         }
     }
+
+    /// Obtains the tile data present at the position given relative to the chunk's top-left-most tile.
+    ///
+    /// If the position given is invalid or the position is empty, this function will return [`None`].
+    ///
+    /// If you want to get a [`LayerTile`](`crate::LayerTile`) instead, use [`ChunkWrapper::get_tile()`].
+    pub fn get_tile_data(&self, x: i32, y: i32) -> Option<&LayerTileData> {
+        if x < Self::WIDTH as i32 && y < Self::HEIGHT as i32 && x >= 0 && y >= 0 {
```
bjorn, line 134: Just wondering, but how useful are these conditions at this low level, given that the main use-case will be iterating the entire chunk as in the example in the docs, making them just a potential performance overhead? Would a debug-only out of bounds check make sense? Of course same consideration applies to fixed tile layers.

src/layers/tile/infinite.rs
```diff
@@ -167,4 +215,51 @@ impl<'map> InfiniteTileLayer<'map> {
...
+    /// # }
+    /// ```
+    #[inline]
+    pub fn chunks(&self) -> impl ExactSizeIterator<Item = ((i32, i32), ChunkWrapper<'map>)> + 'map {
+        let map: &'map crate::Map = self.map;
+        self.data
+            .chunks
+            .iter()
+            .map(move |(pos, chunk)| (*pos, ChunkWrapper::new(map, chunk)))
+    }
```
bjorn, line 264: I guess that a `get_chunk(&self, x: i32, y: i32) -> Option<ChunkWrapper<'map>>` would also be nice, for those who want to render only the visible chunks. Though, I guess there would need to be quite a large number of chunks before this is faster than filtering the iterator returned by `chunks`.

bjorn (approved):
Looks great, though I did suggest some small changes.

src/layers/tile/infinite.rs
aleokdev: An unsafe version of the function could probably be supplied as well, `get_tile_data_unchecked`. I agree.

aleokdev: It is, most containers (e.g. all std ones) have unchecked unsafe functions to obtain data the user is sure exists.

aleokdev: Leaving that for a future PR.

bjorn (approved):

PR #222, opened 2022-05-02
Merge current into next

aleokdev:


bjorn (approved):

bjorn:
Whoops, a branch merge like this should not have been committed as squashed. :-(

I was quite confused why GitHub was still indicating `next` was "7 commits behind `current`".

aleokdev:
Any way to fix it now?

bjorn:
> Any way to fix it now?

We can either "rewrite history", by reverting the `next` branch to before that change and then properly merging in `current`, and repeating the changes done afterwards, or we can leave the mistake be and just follow it up with a proper merge (to make the `next` branch appear up-to-date with the changes in `current`).

Personally, I think rewriting history might be preferable, because if we don't do that the history will just look confusing.

bjorn:
@aleokdev Alright, the `next` branch now contains a proper merge roughly at the same time you did it, and then I've re-applied the commits for #170 and #161. Now, we can try merging `current` into `next` again for change 672b5cab6934b4d7159a729d1c4ad9d6e6acaf2c. It has some conflicts in `object.rs`.

If you have local branches that were based on the previous `next`, you'll need to rebase them before opening a pull request.

bjorn:
@aleokdev Sorry, I forgot about the `LayerType::TileLayer` rename, so I've adjusted the merge again and re-applied the other changes again. Should be all good this time!

PR #224, opened 2022-06-05
remove parent dir on external templates

Aposhian:
This fixes external template loading. I'm not sure if there is another case that this isn't handling though.

### DIFF
--- a/src/objects.rs
+++ b/src/objects.rs
@@ -211,7 +211,10 @@ impl ObjectData {
         // If the template attribute is there, we need to go fetch the template file
         let template = template
             .map(|template_path: String| {
-                let parent_dir = if std::fs::metadata(base_path).or(Err(Error::PathIsNotFile))?.is_file() {
+                let parent_dir = if std::fs::metadata(base_path)
+                    .or(Err(Error::PathIsNotFile))?
+                    .is_file()
+                {
                     base_path.parent().ok_or(Error::PathIsNotFile)?
                 } else {
                     base_path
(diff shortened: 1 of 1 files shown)


aleokdev:
That's strange. By the looks of it, base_path should be set to the path of the file being loaded. Is it the directory it is in instead? (Also, tests seem to be failing, so the issue is probably deeper than this.)

Aposhian:
Yeah, so this fix looks like it isn't quite right: yes base_path was being set to the directory, not the path of the map file.

Aposhian:
I think it is occurring because the template parsing function is being called recursively for some reason. One solution could be to only strip to the parent path if the path is actually a file name?

aleokdev:
Could you send a set of files that reproduces the error? I can work on it this evening.

Aposhian:
Yeah this can be reproduced with the tiled maps at https://github.com/Aposhian/bevy-sandbox/tree/tilemap-refactor/assets

Aposhian:
I just pushed a more complete workaround that both supports external templates and object layers. It may still be a symptomatic fix, but it is working for now for me.

aleokdev:
Sorry for the wait! After some investigation, `base_path` can be *both* a file or a directory :) Somehow that got through all the reviews done to the templates PR.

aleokdev:
@Aposhian I'm working on a proper internal fix for all inconsistencies between file/directory paths. Is it okay if we use a modified version of your map without texture files to test against this particular error?

Aposhian:
Of course! The example repo I linked to is MIT licensed, including all of the assets, which I created.

aleokdev:
Fixed in #227. Thanks for your help!

PR #225, opened 2022-06-12
Initial Wang Set supports

josebaMdeL:


aleokdev:
Oh wow! That's a lot of progress :) Thanks for the PR!

Some notes:
- Is this PR finished? If so, then please add a description of the changes, otherwise mark it as a draft (It's on the "Reviewers" tab on the right)
- Where's `assets/Land1.png` from? It looks like it's from some version of RPG Maker - If so, then we probably can't have it in the repo. Otherwise, please add credit where appropiate, thanks!

josebaMdeL:
Made by mistake!

I thought I was pushing it to my branch; but I'm still testing it.

PR #226, opened 2022-06-13
Initial support for WangSets

josebaMdeL:
### Initial support for WangSets. 

I have added it by reading the TMX Map Format documentation and following some examples, but note that:
1. My code doesn't feet too well on the current crate organization (I will need some help with this).
2. I don't really know how WangSets work or how are they used inside Tiled (I would like someone who knows them to test it).
3.  I don't really know what I'm doing in general (no way to fix this at the moment ;) )


bjorn:
Just a small bit of feedback from me for now, I would suggest you turn your Wang set example into a test instead, since it demonstrates nothing new as an example, but it would be good to cover the loading of Wang sets with a test.

josebaMdeL:
> Just a small bit of feedback from me for now, I would suggest you turn your Wang set example into a test instead, since it demonstrates nothing new as an example, but it would be good to cover the loading of Wang sets with a test.

Done!

src/tileset/wangset.rs
```diff
@@ -0,0 +1,101 @@
...
+    util::{get_attrs, parse_tag, XmlEventResult},
+    Result, TileId,
+};
+
+mod wang_color;
+pub use wang_color::WangColor;
+mod wang_tile;
+pub use wang_tile::{WangId, WangTile};
+
+/// Undocummented WangSet types
```
aleokdev: What does this mean?

src/tileset/wangset.rs
```diff
@@ -0,0 +1,101 @@
...
+
+/// Raw data belonging to a WangSet.
+#[derive(Debug, PartialEq, Clone)]
+pub struct WangSet {
+    /// The name of the Wang set
+    pub name: String,
+    /// Type of wangset
+    pub wang_type: WangType,
+    /// The tile ID of the tile representing this Wang set.
+    pub tile: Option<TileId>,
```
aleokdev, line 39: Assuming this is correct, but just to be sure; @bjorn This is referring to a local ID in the tileset and not a GID right?

src/tileset/wangset/wang_color.rs
```diff
@@ -0,0 +1,63 @@
...
+    util::{get_attrs, parse_tag, XmlEventResult},
+    Result, TileId,
+};
+
+/// Raw data belonging to a tile.
+#[derive(Debug, PartialEq, Clone)]
+pub struct WangColor {
+    /// The name of this color.
+    pub name: String,
+    /// The color in #RRGGBB format (example: #c17d11).
```
aleokdev: The "format" part is not really applicable here. Since this member is pretty self-explanatory I'd just remove the doc comment altogether.
```suggestion
    #[allow(missing_docs)]
```

aleokdev (commented):
Looks good! I'm going to need @bjorn to proofread this though as I'm not very experienced with wang sets.

src/tileset/wangset.rs
bjorn: Yes, since Wang sets are stored as part of the tileset, these tile references use the local IDs.

src/tileset.rs
```diff
@@ -53,6 +56,9 @@ pub struct Tileset {
     /// All the tiles present in this tileset, indexed by their local IDs.
     tiles: HashMap<TileId, TileData>,
 
+    /// All the wangsets present in this tileset
+    pub wangsets: Vec<WangSet>,
```
bjorn: I'd suggest to rename to `wangSets`, for consistency. Or.. I see that'd have to be `wang_sets`, so maybe we leave it be. :/

src/tileset/wangset.rs
```diff
@@ -0,0 +1,101 @@
...
+};
+
+mod wang_color;
+pub use wang_color::WangColor;
+mod wang_tile;
+pub use wang_tile::{WangId, WangTile};
+
+/// Undocummented WangSet types
+#[derive(Debug, PartialEq, Clone, Copy)]
+pub enum WangType {
```
bjorn: It's technically `WangSetType`, but I see this gets a bit ugly with Rust's underscore naming style for variables. Would `wangset_type` be suitable as the variable name, then?

src/tileset/wangset.rs
```diff
@@ -0,0 +1,101 @@
...
+#[derive(Debug, PartialEq, Clone)]
+pub struct WangSet {
+    /// The name of the Wang set
+    pub name: String,
+    /// Type of wangset
+    pub wang_type: WangType,
+    /// The tile ID of the tile representing this Wang set.
+    pub tile: Option<TileId>,
+    /// A color that can be used to define the corner and/or edge of a Wang tile.
+    pub wang_color: Vec<WangColor>,
```
bjorn: Shouldn't that be plural?

src/tileset/wangset/wang_tile.rs
```diff
@@ -0,0 +1,64 @@
...
+use xml::attribute::OwnedAttribute;
+
+use crate::{
+    error::Error,
+    util::{get_attrs, XmlEventResult},
+    Result, TileId,
+};
+
+/**
+The Wang ID, given by a comma-separated list of indexes (starting from 1, because 0 means _unset_) referring to the Wang colors in the Wang set in the following order: top, top right, right, bottom right, bottom, bottom left, left, top left (since Tiled 1.5). Before Tiled 1.5, the Wang ID was saved as a 32-bit unsigned integer stored in the format 0xCECECECE (where each C is a corner color and each E is an edge color, in reverse order).
```
bjorn: This documentation should be adjusted to describe what the Rust structure means, rather than talking about the various ways it is or was stored in the TMX format.

src/tileset/wangset/wang_tile.rs
```diff
@@ -0,0 +1,64 @@
...
+        }
+        for i in 0..8 {
+            ret[i] = s[i].parse::<u32>().unwrap_or(0);
+        }
+
+        Ok(WangId(ret))
+    }
+}
+
+/// Raw data belonging to a tile.
```
bjorn: Description seems unsuitable. Maybe rather something like "Stores a tile reference along with its associated Wang ID."

bjorn (commented):
Some comments, mostly about naming and documentation.

src/tileset.rs
aleokdev: Can't be wangSets because of standard formatting rules. wang_sets could be ok.

src/tileset/wangset.rs
bjorn: Ok, let's be consistent and go with `WangSet` and `wang_set`, and thus here `WangSetType` and `wang_set_type`.

src/tileset/wangset/wang_tile.rs
```diff
@@ -0,0 +1,64 @@
...
+};
+
+/**
+The Wang ID, given by a comma-separated list of indexes (starting from 1, because 0 means _unset_) referring to the Wang colors in the Wang set in the following order: top, top right, right, bottom right, bottom, bottom left, left, top left (since Tiled 1.5). Before Tiled 1.5, the Wang ID was saved as a 32-bit unsigned integer stored in the format 0xCECECECE (where each C is a corner color and each E is an edge color, in reverse order).
+*/
+#[derive(Debug, PartialEq, Clone, Copy)]
+pub struct WangId(pub [u32; 8]);
+
+impl FromStr for WangId {
+    type Err = ();
```
aleokdev: You probably saw this on some other code from the crate too, but it's not good practice to set the error to the unit type. I'd prefer if you made a unit struct `struct WangIdParseError;` which implemented `std::error::Error` instead

josebaMdeL: Even better, should I use type Err = Error; from crate::error::Error? I could also add a new type of error on Error enum.
For example:
```rust
/// WangId not properly formated. Stores the wrongly parsed String.
InvalidWangIdEncoding(String),

/// This way we can later display error like:
write!(fmt, "\"{}\" Is not a valid WangId format", s),`

File: src/tileset/wangset.rs (hunk 6 above)
Comment (josebaMdeL) [reply src/tileset/wangset.rs]: Yes!

File: src/tileset/wangset.rs (hunk 1 above)
Comment (josebaMdeL) [reply src/tileset/wangset.rs]: That wangset types are not described on Tiled documentation, but they appear (and must be chosen) on the files. I didn't know how to describe them as I don't know what the differences are.

Comment (bjorn) [reply src/tileset/wangset.rs]: Whoops, this should be fixed in the TMX format docs, which indeed forgets to mention the type entirely at https://doc.mapeditor.org/en/stable/reference/tmx-map-format/#wangset.

For the user, the various types are documented at https://doc.mapeditor.org/en/stable/manual/terrain/.

File: src/tileset/wangset/wang_tile.rs (hunk 9 above)
Comment (aleokdev) [reply src/tileset/wangset/wang_tile.rs]: If it is an error that can be thrown when loading maps, sure.

**Comment by josebaMdeL** [comment]
Updated with proposed changes. I just forget to change the WangSetType's doc comment, but it will be corrected for the next commit.

File: src/error.rs [hunk 10]
```diff
@@ -49,6 +49,8 @@ pub enum Error {
         /// Supported types are `string`, `int`, `float`, `bool`, `color`, `file` and `object`.
         type_name: String,
     },
+    /// WangId not properly formated. Stores the wrongly parsed String.
+    InvalidWangIdEncoding(String),
```
aleokdev: This member should be named instead, a String is not very informative of what it represents in this context.

src/tileset/wangset.rs
```diff
@@ -0,0 +1,101 @@
...
+    }
+}
+
+/// Raw data belonging to a WangSet.
+#[derive(Debug, PartialEq, Clone)]
+pub struct WangSet {
+    /// The name of the Wang set
+    pub name: String,
+    /// Type of wangset
+    pub wang_type: WangSetType,
```
bjorn: Let's go with `wang_set_type` for consistency.

src/tileset/wangset.rs
```diff
@@ -0,0 +1,101 @@
...
+    /// The tile ID of the tile representing this Wang set.
+    pub tile: Option<TileId>,
+    /// A color that can be used to define the corner and/or edge of a Wang tile.
+    pub wang_colors: Vec<WangColor>,
+    /// A color that can be used to define the corner and/or edge of a Wang tile.
+    pub wang_tiles: HashMap<TileId, WangTile>,
+    /// The custom properties of this tile.
```
bjorn:
```suggestion
    /// The custom properties of this Wang set.
```

src/tileset/wangset.rs
```diff
@@ -0,0 +1,101 @@
...
+        WangSetType::Mixed
+    }
+}
+
+/// Raw data belonging to a WangSet.
+#[derive(Debug, PartialEq, Clone)]
+pub struct WangSet {
+    /// The name of the Wang set
+    pub name: String,
+    /// Type of wangset
```
bjorn:
```suggestion
    /// Type of Wang set
```

src/tileset/wangset/wang_color.rs
```diff
@@ -0,0 +1,63 @@
...
+pub struct WangColor {
+    /// The name of this color.
+    pub name: String,
+    #[allow(missing_docs)]
+    pub color: Color,
+    /// The tile ID of the tile representing this color.
+    pub tile: Option<TileId>,
+    /// The relative probability that this color is chosen over others in case of multiple options. (defaults to 0)
+    pub probability: f32,
+    /// The custom properties of this tile.
```
bjorn:
```suggestion
    /// The custom properties of this color.
```

src/tileset/wangset.rs
```diff
@@ -0,0 +1,101 @@
...
+    pub wang_type: WangSetType,
+    /// The tile ID of the tile representing this Wang set.
+    pub tile: Option<TileId>,
+    /// A color that can be used to define the corner and/or edge of a Wang tile.
+    pub wang_colors: Vec<WangColor>,
+    /// A color that can be used to define the corner and/or edge of a Wang tile.
+    pub wang_tiles: HashMap<TileId, WangTile>,
```
bjorn, line 43: I wonder if this couldn't be just `HashMap<TileId, WangId>`, since the `WangTile` is just going to contain the same `TileId` again, right?

We might not need a `WangTile` struct at all. In Tiled it only exists temporarily as a way to prepare the data for saving in a consistent order.

The documentation string still needs to be adjusted.

src/tileset/wangset.rs
```diff
@@ -0,0 +1,101 @@
...
+/// Raw data belonging to a WangSet.
+#[derive(Debug, PartialEq, Clone)]
+pub struct WangSet {
+    /// The name of the Wang set
+    pub name: String,
+    /// Type of wangset
+    pub wang_type: WangSetType,
+    /// The tile ID of the tile representing this Wang set.
+    pub tile: Option<TileId>,
+    /// A color that can be used to define the corner and/or edge of a Wang tile.
```
bjorn:
```suggestion
    /// The colors that can be used to define the corner and/or edge of each Wang tile.
```

bjorn (commented):
A few more comments.

src/tileset/wangset.rs
josebaMdeL: Well, WangTile describes the tile connections that a single Tile has on that Wang Set. If you delete that 8 size array, not sure you can get that data again. The only simplification I see we could do is to delete the TileId value that is hold inside WangTile struct, as it can be recovered from the wang_tiles HashMap. I just added there as it's very small in size and can be useful to have.

bjorn: Yes, the `WangTile` is just tile ID + `WangId` currently. If you remove tile ID, all that left is the `WangId`, in which case why would you have a wrapping `WangTile` struct at all? That's why I was suggesting to remove it.

josebaMdeL: Well, I would prefer to keep the wrapping struct just to implement FromStr and be able to return an encoding error. Seems more clear to me.

Also, if there is more data inserted on this field in the future, we already have the struct prepared.

src/error.rs
josebaMdeL: Something like:
```rust
    /// WangId not properly formated.
    InvalidWangIdEncoding {
        /// Stores the wrongly parsed String.
        readed_string: String,
    },
`

File: src/tileset/wangset.rs (hunk 15 above)
Comment (bjorn) [reply src/tileset/wangset.rs:43]: Alright, we can keep the struct if that is more convenient, but do we need to actually use it in the `wang_tiles` hash map, or could we keep the `WangTile` for parsing and error reporting, but store just the `WangId` there? In general in other hash maps, we have avoided storing the key as part of the value as well.

File: src/error.rs (hunk 10 above)
Comment (aleokdev) [reply src/error.rs]: Sure, but it should be "formatted" and "read" instead of "readed".

File: src/tileset/wangset.rs (hunk 15 above)
Comment (josebaMdeL) [reply src/tileset/wangset.rs:43]: Leaving like this?:

```rust
pub struct WangTile {
    /// The Wang ID,
    pub wang_id: WangId,
}
`

Comment (aleokdev) [reply src/tileset/wangset.rs:43]: > Well, I would prefer to keep the wrapping struct just to implement FromStr and be able to return an encoding error. Seems more clear to me.
> 
> Also, if there is more data inserted on this field in the future, we already have the struct prepared.

Seems wiser to make the hashmap private, change the value type and expose an access method that returns the wrapping struct, as done everywhere else in the crate (i.e. Data vs wrapper structures)

Comment (aleokdev) [reply src/tileset/wangset.rs:43]: > Alright, we can keep the struct if that is more convenient, but do we need to actually use it in the `wang_tiles` hash map, or could we keep the `WangTile` for parsing and error reporting, but store just the `WangId` there? In general in other hash maps, we have avoided storing the key as part of the value as well.

Exactly this, as done with tilesets & their stored tiles, for instance. Store raw/minimal data in hashmap, return more convenient wrapper to user.

Comment (josebaMdeL) [reply src/tileset/wangset.rs:43]: > 

About this, I don't see any advantage on returning, for example, an iterator in the same way that we return tiles in Tileset, as no new info will be added to WangTile struct.

File: src/error.rs [hunk 17]
```diff
@@ -93,6 +98,8 @@ impl fmt::Display for Error {
                 write!(fmt, "Invalid property value: {}", description),
             Error::UnknownPropertyType { type_name } =>
                 write!(fmt, "Unknown property value type '{}'", type_name),
+            Error::InvalidWangIdEncoding{read_string} =>
+                write!(fmt, "\"{}\" Is not a valid WangId format", read_string),
```
aleokdev:
```suggestion
                write!(fmt, "\"{}\" is not a valid WangId format", read_string),
```

src/error.rs
```diff
@@ -49,6 +49,11 @@ pub enum Error {
         /// Supported types are `string`, `int`, `float`, `bool`, `color`, `file` and `object`.
         type_name: String,
     },
+    /// WangId not properly formatted.
```
aleokdev:
```suggestion
    /// Found a WangId that was not properly formatted.
```

src/tileset.rs
```diff
@@ -53,6 +56,9 @@ pub struct Tileset {
     /// All the tiles present in this tileset, indexed by their local IDs.
     tiles: HashMap<TileId, TileData>,
 
+    /// All the wangsets present in this tileset
```
aleokdev:
```suggestion
    /// All the wangsets present in this tileset.
```

src/tileset/wangset/wang_tile.rs
```diff
@@ -0,0 +1,64 @@
...
+        }
+        for i in 0..8 {
+            ret[i] = values[i].parse::<u32>().unwrap_or(0);
+        }
+
+        Ok(WangId(ret))
+    }
+}
+
+/// Stores a tile reference along with its associated Wang ID.
```
aleokdev: The docstring is not accurate.

src/tileset/wangset/wang_tile.rs
```diff
@@ -0,0 +1,64 @@
...
+#[derive(Debug, PartialEq, Clone, Copy)]
+pub struct WangTile {
+    /// The Wang ID,
```
aleokdev:
```suggestion
#[allow(missing_docs)]
```
This doc comment doesn't contribute much so I'd say to get rid of it.

src/tileset/wangset/wang_tile.rs
```diff
@@ -0,0 +1,64 @@
...
+    pub wang_id: WangId,
+}
+
+impl WangTile {
+    /// Reads data from XML parser to create a WangTile.
+    pub fn new(
```
aleokdev:
```suggestion
    pub(crate) fn new(
```
Parsing functions are set to pub(crate), or at least for now. Let's change it for consistency.

aleokdev (commented):
Sorry for the delay. This should be the final review, I'm OK with this being on the next 0.10 release.

src/tileset/wangset/wang_tile.rs
josebaMdeL: Hi, could I just commit al the changes in a single commit here from Github?, I see I can rapidly commit every change, but one by one.

aleokdev: I think there's a way to stage changes and then push them all at once but I'm not sure how to do it, sorry

aleokdev (dismissed):
LGTM.

src/tileset/wangset/wang_tile.rs
```diff
@@ -0,0 +1,64 @@
...
+    error::Error,
+    util::{get_attrs, XmlEventResult},
+    Result, TileId,
+};
+
+/**
+The Wang ID, stored as an array of 8 u32 values.
+*/
+#[derive(Debug, PartialEq, Clone, Copy)]
+pub struct WangId(pub [u32; 8]);
```
bjorn: I wonder a bit why `u32` is used here instead of `u8`. In Tiled, a `WangId` is actually stored as a single `u64`, but storing as 8 times `u8` would also make sense. It's just that some operations, like masking, are easier to do on a single value than on an array of bytes.

bjorn (dismissed):
Looks fine to me as well, though I left one comment that I was wondering about.

src/tileset/wangset/wang_tile.rs
josebaMdeL: You are right, I see now in the docs that this field should be u8 values instead of u32.

I was mistaken because this was not specified on TMX format docs, but it is in JSON.

> wangid: "The Wang ID, given by a comma-separated list of indexes (starting from 1, because 0 means _unset_) referring to the Wang colors in the Wang set in the following order: top, top right, right, bottom right, bottom, bottom left, left, top left (since Tiled 1.5). Before Tiled 1.5, the Wang ID was saved as a 32-bit unsigned integer stored in the format 0xCECECECE (where each C is a corner color and each E is an edge color, in reverse order)."

bjorn (dismissed):
Looks great, thanks a lot for your work on this!

src/tileset/wangset/wang_tile.rs
bjorn: Thanks for pointing that out, I'll push a small update to those docs to mention the range of the indexes (0-255).

bjorn (approved):

PR #227, opened 2022-06-13
Fix & add tests for template object loading

aleokdev:
Fixes the underlying issue behind #224, and adds a few tests with a map provided by @Aposhian (Thank you!). It also makes `DefaultResourceCache`'s members public for easy access from tests and whatnot.


src/objects.rs
```diff
@@ -211,8 +217,15 @@ impl ObjectData {
         // If the template attribute is there, we need to go fetch the template file
         let template = template
             .map(|template_path: String| {
-                let parent_dir = base_path.parent().ok_or(Error::PathIsNotFile)?;
-                let template_path = parent_dir.join(Path::new(&template_path));
+                // The base path may be a directory or a file
```
bjorn: Does this added comment refer to the removed `base_path`? It's outdated, right?

src/objects.rs
```diff
@@ -211,8 +217,15 @@ impl ObjectData {
...
+                let template_path = {
+                    let path = PathBuf::from(template_path);
+                    if path.is_relative() {
+                        file_path_being_loaded.parent().unwrap().join(&path)
```
bjorn: Might there be a performance penalty from using `parent().unwrap()` in potentially many places? I personally liked that we were passing around the directory instead of the file path.

bjorn (commented):
What caused the actual inconsistency in the base path?

src/objects.rs
aleokdev: Yes, it's outdated

src/objects.rs
aleokdev: Rather quite the opposite, we were previously prematurely obtaining the directory of the file being loaded before. Now, we only obtain that directory when it is actually required (when doing path manipulation for relative paths).

aleokdev: Plus, ensuring that we are passing a file path as opposed to a file or directory path allows us to completely remove the "Path does not refer to a file" error in many places.

aleokdev:
> What caused the actual inconsistency in the base path?

It was being provided both file and directory paths while the code inside assumed it was a directory path only. By making the name clearer I think we'll avoid other issues like this one.

src/objects.rs
bjorn: > Rather quite the opposite, we were previously prematurely obtaining the directory of the file being loaded before. Now, we only obtain that directory when it is actually required (when doing path manipulation for relative paths).

Right, but previously it could happen exactly once per file, whereas now it happens once per file reference. Generally one can expect there to be a lot more file references than files, especially in template-heavy maps.

> Plus, ensuring that we are passing a file path as opposed to a file or directory path allows us to completely remove the "Path does not refer to a file" error in many places.

Yeah, I understand we need this to be consistent. I was just wondering whether it would be better to consistently pass the parent directory rather than consistently using the file path. Anyway, I also doubt it really matters.

bjorn:
> It was being provided both file and directory paths while the code inside assumed it was a directory path only.

Yes, but in my quick review I didn't notice the location where we were passing a file instead of a directory path. I'd personally prefer a patch that just fixes that location.

src/layers/mod.rs
```diff
@@ -102,7 +102,7 @@ impl LayerData {
                     attrs,
                     Some(tilesets),
                     for_tileset,
-                    map_path,
+                    map_path.parent().ok_or(crate::Error::PathIsNotFile)?,
```
bjorn, line 105: Could we pass in the parent directory to `LayerData::new` in the first place? Or does the `map_path` parameter to `ImageLayerData::new` and `GroupLayerData::new` need to be the actual map path?

aleokdev: We could change that, but I was trying to change as few lines as possible since you asked to patch just the bug.

bjorn: Yeah, that's appreciated! It just highlighted that the inconsistency goes further than this one line. But, we can merge this fix and make it consistent in a follow-up change.

For following this up, I think we should always pass just the directory.

bjorn (approved):

PR #229, opened 2022-09-06
Merge current into next

aleokdev:


aleokdev:
Also, maybe it'd be a good idea to close any further PRs for features for current, and focus only on `next`, so that we can move all the XML parsing functions into the `parse` folder, which is currently pretty empty:
Screenshot_20220906_175223

bjorn (approved):
Thanks for taking care of this!

PR #230, opened 2022-09-06
Add text objects

aleokdev:
How didn't we have these?? Major oversight :)

Merge after #229.

### DIFF
--- a/src/objects.rs
+++ b/src/objects.rs
@@ -7,7 +7,7 @@ use crate::{
     properties::{parse_properties, Properties},
     template::Template,
     util::{get_attrs, map_wrapper, parse_tag, XmlEventResult},
-    Gid, MapTilesetGid, ResourceCache, ResourceReader, Tile, TileId, Tileset,
+    Color, Gid, MapTilesetGid, ResourceCache, ResourceReader, Tile, TileId, Tileset,
 };
 
 /// The location of the tileset this tile is in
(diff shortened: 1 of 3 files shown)


bjorn:
I'd suggest to rebase this PR now.

src/objects.rs
```diff
@@ -333,6 +381,95 @@ impl ObjectData {
...
+                color,
+                bold,
+                italic,
+                underline,
+                strikeout,
+                kerning,
+                halign,
+                valign,
+
+        )
```
bjorn: Code formatting seems a bit strange here.

src/objects.rs
```diff
@@ -333,6 +381,95 @@ impl ObjectData {
...
+        let font_family = font_family.unwrap_or_else(|| "sans-serif".to_string());
+        let pixel_size = pixel_size.unwrap_or(16);
+        let wrap = if wrap == Some(1) { true } else { false };
+        let color = color.unwrap_or(Color {
+            red: 0,
+            green: 0,
+            blue: 0,
+            alpha: 255,
+        });
+        let bold = if bold == Some(1) { true } else { false };
```
bjorn: Why do we need `let bold = if bold == Some(1) { true } else { false }`? Could just `let bold = bold == Some(1)` suffice? Or `bold.contains(&1)`, heh.

bjorn (commented):
Looks good, just two small comments.

src/objects.rs
aleokdev: Oh lol yeah

aleokdev:
Contains is unstable so this will have to do :)

bjorn (dismissed):

bjorn:
Approved, but don't forget the changelog. :-)

bjorn (approved):

PR #232, opened 2022-09-07
Move parsing functions to parse module

aleokdev:
WIP. This PR moves all internal functions to the parse module.
This has the advantage of providing a clear interface to the user & developers, as the structure of Tiled objects are now defined completely separate from how they are parsed. However, this PR still doesn't answer questions like "Should we make the parse functions public?" "How should we present the parse functions internally?" etc. For now the definitions have been moved literally to the XML/"common" module.

Stopped mid-way since I realized a huge mistake with wangsets (The `WangSetType` type isn't accessible from outside and the compiler didn't complain somehow, even with the `wang_set_type` public WangSet member). Fixing this now in `current`. I wish we'd have implemented wangsets in `next` instead. Oh well.


aleokdev:
This is pretty outdated so I'll close it for now. I might do this at some point, just need to have time and patience...

PR #238, opened 2022-12-07
Support Tiled 1.9 `class` property

### LINKED ISSUE #234: since tiled 1.9 object type is renamed into class
As written in https://www.mapeditor.org/2022/06/25/tiled-1-9-released.html

> Unified Custom Types
> 
> Object Types have been merged into the Property Types, now simply called “Custom Types”. This means custom classes can now have a color and can be used as “object type”. If your project refers to an object types file, these types will be automatically imported as custom classes. If you used globally stored object types before, they can be manually imported to your project.
> 
> In addition, the “Type” property previously available only for objects and tiles is now available for all data types as the new “Class” property. For consistency, this value is written out as “class” also for objects and tiles, but a project-wide compatibility option is provided to make it still write out as “type”:

Maybe rs-tiled should support the `class` attribute by default, or just automatically read the class into obj_type.
Anyhow for now I will use the compatibility option of tiled

barthap:
### Why

Closes #234 

In Tiled 1.9 the "type" property was renamed to "class". My PR addresses this.

### How

In addition to "type", the "class" property is parsed. If the former is not present, the latter takes its place. This way, the solution is compatible with both 1.9+ and older maps. 

After the below conversation, this is what I did:
- Deprecated `obj_type` for Objects and `tile_type` for Tiles in favor of `user_type`. Added deprecation notices
- Added `user_type` to Maps, Tilesets and Layers.
- Both `class` and `type` XML params now resolve to `user_type` everywhere.

Support for type/class is now implemented for the following entities:
- Tiles in tileset (`Tile`)
- Objects
- Maps
- Tilesets
- Layers

These are all places that I'm aware of.

### Test plan

Tested locally on both old and new maps.
I didn't add tests as I haven't found any existing ones (nor test fixtures) which have the "type" or "class" property set. Eventually, I can create/modify one if requested.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [Unreleased]
 ### Added
 - Support for Wang sets.
+- Support for Tiled 1.9+ `class` property.
 
 ### Changed
 - Update `zstd` to `0.12.0`.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [Unreleased]
 ### Added
 - Support for Wang sets.
-- Support for Tiled 1.9+ `class` property.
+- Support for Tiled 1.9 `Class` property. Maps, tilesets and layers now have a `user_type` property. 
+
+### Deprecated
+- `Tile::tile_type`: Use `Tile::user_type` instead.
+- `Object::obj_type` Use `Object::user_type` instead.
 
 ### Changed
 - Update `zstd` to `0.12.0`.
--- a/src/objects.rs
+++ b/src/objects.rs
@@ -34,7 +34,7 @@ pub struct ObjectData {
     /// The type of the object, which is arbitrary and set by the user.
     pub user_type: String,
     /// This property has been renamed to `user_type`.
-    #[deprecated(since = "0.11.0", note = "Use [`ObjectData::user_type`] instead")]
+    #[deprecated(since = "0.10.3", note = "Use [`ObjectData::user_type`] instead")]
     pub obj_type: String,
     /// The width of the object, if applicable. This refers to the attribute in `object`.
     /// Since it is duplicate or irrelevant information in all cases, use the equivalent
--- a/src/objects.rs
+++ b/src/objects.rs
@@ -32,6 +32,9 @@ pub struct ObjectData {
     /// The name of the object, which is arbitrary and set by the user.
     pub name: String,
     /// The type of the object, which is arbitrary and set by the user.
+    pub user_type: String,
+    /// This property has been renamed to `user_type`.
+    #[deprecated(since = "0.11.0", note = "Use [`ObjectData::user_type`] instead")]
     pub obj_type: String,
     /// The width of the object, if applicable. This refers to the attribute in `object`.
     /// Since it is duplicate or irrelevant information in all cases, use the equivalent
@@ -86,8 +89,8 @@ impl ObjectData {
                 Some("id") => id ?= v.parse(),
                 Some("gid") => tile ?= v.parse(),
                 Some("name") => name ?= v.parse(),
-                Some("type") => obj_type ?= v.parse(),
-                Some("class") => obj_class ?= v.parse(),
+                Some("type") => user_type ?= v.parse(),
+                Some("class") => user_class ?= v.parse(),
                 Some("width") => width ?= v.parse(),
                 Some("height") => height ?= v.parse(),
                 Some("visible") => visible ?= v.parse().map(|x:i32| x == 1),
@@ -96,9 +99,8 @@ impl ObjectData {
                 "x" => x ?= v.parse::<f32>(),
                 "y" => y ?= v.parse::<f32>(),
             }
-            ((id, tile, name, obj_type, obj_class, width, height, visible, rotation), (x, y))
+            ((id, tile, name, user_type, user_class, width, height, visible, rotation), (x, y))
         );
-        let t = t.or(c);
         let tile = tile.and_then(|bits| LayerTileData::from_bits(bits, tilesets?));
         let visible = v.unwrap_or(true);
         let width = w.unwrap_or(0f32);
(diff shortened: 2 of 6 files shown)


aleokdev:
Thank you! Could you also update the changelog (On the "Unreleased" section)?

barthap:
~Ouch, actually skipping CI blocked the required workflow from running 🙄 Force-push might help~ Fixed

aleokdev:
@bjorn Could you check this out? I'm still not 100% familiar with Tiled 1.9 changes

barthap:
Actually, I found out, that layers (all kinds of them?) and maps can have a `class` property too.

So my PR only fixes this for existing entities which previously had `type`. I can add it to the others, but how the property should be named? I thought about `map_type` etc.. to keep the convention, but for layers there is already a (private) `layer_type` so I'm not sure how to name it. I don't know your breaking change policy, but maybe let's change this to property named `{obj,map,tileset,layer}_class: String`? 🤔

aleokdev:
Bjorn plans to rename `class` back to `type` on 1.10, so I don't think the rename is worth it:

> [...] that rename didn't go very well, everybody hates their project breaking due to some arbitrary format change. So I plan to rename "class" back to "type", [...]

barthap:
> Bjorn plans to rename class back to type on 1.10, so I don't think the rename is worth it

Ok, in this case it would be good to just add support for the type/class property to all entities, without renaming it. I pushed a proposal commit.

bjorn:
If we go for `user_type` then I think we can consistently use this name rather than `map_type` and `tileset_type` and such. I realize we already have `obj_type`, which made sense when it only applied to objects, but maybe it could be renamed to `user_type` as well now.

And indeed, I plan to rename `class` back to `type` for Tiled 1.10, in which case it would be `type` everywhere else too. While we're at it, it might be good to already support this.

barthap:
Yes, in this case, using the `user_type` everywhere makes the most sense. I've updated my PR to:
- Deprecated `obj_type` for Objects and `tile_type` for Tiles in favor of `user_type`. Added deprecation notices
- Added `user_type` to Maps, Tilesets and Layers.
- Both `class` and `type` XML params now resolve to `user_type` everywhere.

I have found one inconsistency. For tiles, the `tile_type` was of type `Option<String>`, and for Objects it was just `String` and optional unwrapped to empty string.
Which way should be preferred? I'd go with the optional type instead of empty strings everywhere, but for now I left it as-is for objects.

src/objects.rs
```diff
@@ -32,6 +32,9 @@ pub struct ObjectData {
     /// The name of the object, which is arbitrary and set by the user.
     pub name: String,
     /// The type of the object, which is arbitrary and set by the user.
+    pub user_type: String,
+    /// This property has been renamed to `user_type`.
+    #[deprecated(since = "0.11.0", note = "Use [`ObjectData::user_type`] instead")]
```
aleokdev: Actually `current` refers to the next 0.10 version, so that would be 0.10.3.

src/tile.rs
```diff
@@ -27,6 +27,9 @@ pub struct TileData {
     /// The animation frames of this tile.
     pub animation: Option<Vec<Frame>>,
     /// The type of this tile.
+    pub user_type: Option<String>,
+    /// This property has been renamed to `user_type`.
+    #[deprecated(since = "0.11.0", note = "Use [`TileData::user_type`] instead")]
```
aleokdev: Same issue here.

src/objects.rs
barthap: Oh, yeah, good catch! I have custom-patched 0.11 in my project and missed that 'current' is at 0.10

aleokdev:
> I have found one inconsistency. For tiles, the `tile_type` was of type `Option<String>`, and for Objects it was just `String` and optional unwrapped to empty string.

This inconsistency also exists in the TMX docs, so I guess we should leave it as is. @bjorn Thoughts?

aleokdev (approved):

aleokdev:
Eh, it's good enough 🙂

aleokdev:
Thank you! 🙏

CHANGELOG.md
```diff
@@ -8,7 +8,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [Unreleased]
 ### Added
 - Support for Wang sets.
-- Support for Tiled 1.9+ `class` property.
+- Support for Tiled 1.9 `Class` property. Maps, tilesets and layers now have a `user_type` property. 
+
+### Deprecated
+- `Tile::tile_type`: Use `Tile::user_type` instead.
+- `Object::obj_type` Use `Object::user_type` instead.
```
bjorn, line 15: When something is "Deprecated" I expect it to still be available, but as far as I understand the patch, these have been "Removed"?

bjorn:
> This inconsistency also exists in the TMX docs, so I guess we should leave it as is. @bjorn Thoughts?

Hmm, I guess the docs have to be fixed. While this property is optional in the file, storage wise I've always defaulted to an empty string rather than having an explicit "type has not been specified" state. Is it helpful to have such a state in Rust?

CHANGELOG.md
aleokdev: Made me pretty worried right there, but no, they haven't been removed but rather marked with the deprecated attribute.

aleokdev:
> Hmm, I guess the docs have to be fixed. While this property is optional in the file, storage wise I've always defaulted to an empty string rather than having an explicit "type has not been specified" state. Is it helpful to have such a state in Rust?

Is it helpful to distinguish between having an empty text attribute vs not having it at all? I don't know 🙂 Depends on the format and editor, not the language really

CHANGELOG.md
barthap: They're still here, but marked as deprecated. I renamed them at first, but then decided to keep old names as "deprecated":
Screenshot 2022-12-16 at 11 08 41

barthap:
Regarding Option vs empty string, I have mixed feelings about that:
- From my perspective, the option seems to be more "correct" from the language point of view, which means the value hasn't been set. An empty string suggests that a value exists. 
- On the other hand, in this particular case it's more convenient to compare against empty string than option:
  ```rust
  // option
  if tile.user_type.as_deref() == Some("light_source") { /* ... */ }

  // empty string
  if tile.user_type == "light_source" { /* ... */ }
  ```

CHANGELOG.md
bjorn: You're right, I must have misinterpreted the diff, sorry.

bjorn:
Right, then I think we should not use `Option`. Since it is both less convenient, and actually in the editor a "type has not been set" state does not exist either, or equals the value being an empty string.

PR #252, opened 2023-02-13
Add zstd WASM support & better docs

### LINKED ISSUE #240: Default zstd feature is unhelpful for wasm builds
Out of the box, adding tiled as a dependency of my project broke my WASM builds because zstd's default build does not support it. While I eventually learned that zstd is an optional feature in tiled, this does not appear to be documented anywhere, and there aren't good interfaces for discovering crate features.

I've seen #228, and I have a couple possibilities to make the default experience better:
* keep zstd as an optional feature but default disabled
* replace zstd with another implementation that supports WASM ([branch that implements this](https://github.com/jdm/rs-tiled/tree/ruzstd))
* expose zstd's `wasm` crate feature somehow and document how to build for WASM? This is the least thought through of the alternatives because I haven't looked into it at all; I just saw that it exists.
* clearly document how to get a wasm-capable build (and the limitations around embedded tilesets), but otherwise make no changes to the status quo

aleokdev:
Closes #240.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 ### Added
+- WASM support + `wasm` feature; Check `README.md` for instructions on how to set it up.
 - Support for staggered maps. Maps now have an `stagger_axis` and `stagger_index` property.
 
 ## [0.11.0]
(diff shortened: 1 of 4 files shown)


bjorn:
What about the `wasm` feature of the `zstd` crate? It's not going to just work?

aleokdev:
It doesn't seem like it from what I've read, but I have yet to test it.

bjorn (approved):
Nice!

PR #254, opened 2023-02-19
Updated lifetime of LayerTile returned from TileLayer

Anti-Alias:
Small PR to change the lifetime specifier of the LayerTile returned by the get_tile methods.
Previously, the LayerTile returned by these methods would have the elided lifetime of "self", which is shorter than 'map.

### DIFF
--- a/src/layers/tile/finite.rs
+++ b/src/layers/tile/finite.rs
@@ -85,7 +85,7 @@ impl<'map> FiniteTileLayer<'map> {
     /// Obtains the tile present at the position given.
     ///
     /// If the position given is invalid or the position is empty, this function will return [`None`].
-    pub fn get_tile(&self, x: i32, y: i32) -> Option<LayerTile> {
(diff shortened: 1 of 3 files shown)


aleokdev:
Argh, not the first time it has happened. Thanks!

aleokdev (approved):

bjorn:
Just wondering, but aren't these the kind of errors the Rust compiler is supposed to be able to notice? Or is this patch triggered by compiler errors, in usage that our tests and examples aren't covering so far?

aleokdev:
Both lifetimes are valid, but 'map lasts longer than 'self. There are no tests for it so the compiler accepts it.

aleokdev:
I forgor to check the changelog again 💀

bjorn:
Ah, so the lifetime was valid, but shorter than it could be, and for convenience we'd want the lifetime to be as long as possible.

aleokdev:
Yup, just like in #201.

PR #272, opened 2023-05-25
Improve ResourceReader's API, add better docs

### LINKED ISSUE #266: More convenient way to load embedded maps
Hi! I'm trying out Bevy and (similar Amethyst) it passes a `&[u8]` to load the file. As of version 0.11, I have no idea how can load maps as implementing `ResourceReader` for loading from byte arrays would have to be intricate to then be able to load the remaining files from disk.

Is there any chance to add back this super useful method? If not, how can I load the main tmx from bytes and the rest of the files using `FilesystemResourceReader`?

aleokdev:
Closes #266. Implements ResourceReader for functions so that you can do stuff like this:
```rust
use tiled::{DefaultResourceCache, Loader};

let mut loader = Loader::with_cache_and_reader(
    DefaultResourceCache::new(),
    // Specify the reader to use. We can use anything that implements `ResourceReader`, e.g. FilesystemResourceReader.
    // Any function that has the same signature as `ResourceReader::read_from` also implements it.
    // Here we define a reader that embeds the map at "assets/tiled_xml.csv" into the executable, and allow
    // accessing it only through "/my-map.tmx"
    // ALL maps, tilesets and templates will be read through this function, even if you don't explicitly load them
    // (They can be dependencies of one you did want to load in the first place).
    // Doing this embedding is useful for places where the OS filesystem is not available (e.g. WASM applications).
    |path: &std::path::Path| -> std::io::Result<_> {
        if path == std::path::Path::new("/my-map.tmx") {
            Ok(std::io::Cursor::new(include_bytes!("../assets/tiled_csv.tmx")))
        } else {
            Err(std::io::ErrorKind::NotFound.into())
        }
    }
);
```
Also adds a FAQ section on loading embedded maps and improves `Loader::with_cache_and_reader` docs.


aleokdev:
~~Not sure why current merge commits are still showing up in the timeline.~~ Edit: Nevermind, fixed

README.md
```diff
@@ -34,7 +34,35 @@ fn main() {
...
+## FAQ
+### How do I embed a map into my executable? / How do I read a file from anywhere else that isn't the filesystem's OS?
+The crate does all of its reading through the `read_from` function of the [`ResourceReader`](https://docs.rs/tiled/latest/tiled/trait.ResourceReader.html) that you create the loader with. By default, this reader is set to [`FilesystemResourceReader`](https://docs.rs/tiled/latest/tiled/struct.FilesystemResourceReader.html) and all files are read through the OS's filesystem. You can however change this.
+
+Here's an example mostly taken from `Loader::with_cache_and_reader`'s documentation:
+```rust
+use tiled::{DefaultResourceCache, Loader};
+
+let mut loader = Loader::with_cache_and_reader(
+    DefaultResourceCache::new(),
```
bjorn: Since you're going for some more convenience with this automatic `ResourceReader` implementation based on a function, maybe it also makes sense to add a `Loader::with_reader` function so that the `DefaultResourceCache::new()` part can be left out? I guess the need to customize the cache could be rare enough?

bjorn (dismissed):
Nice simplification! :-)

README.md
aleokdev: Hmm, true

aleokdev:
Reminder: also change WASM section in readme

bjorn (approved):
Nice!

Issue #166, opened 2022-02-20
Tileset helper for obtaining UVs


alexdevteam:
Add a function for obtaining the UV of a tile given its ID.

alexdevteam:
Perhaps it would be a good idea to also create a `TileData` and make `Tile` an encapsulation, which would avoid the extra index if the user already had the tile in question.

Anti-Alias:
> Perhaps it would be a good idea to also create a TileData and make Tile an encapsulation, which would avoid the extra index if the user already had the tile in question.

I am for this, as it keeps the API _flatter_ and more performant.
It means, excluding ```Tileset``` and maybe some other types, that the user only needs to talk to the facade types rather than sometimes talking to ```Type1<'map> ```and sometimes talking to ```&'map Type2```, which is a little inconsistent.
(Perhaps, if the situation changes, Tileset may need its own wrapper, though I don't know of any usecases yet)

I like your suggestion in #164 of just returning a rect measured in pixels.
I think we can also add a helper function to this new ```Tile<'map>``` type called ```uvs(flip_y: bool) -> ((f32, f32), (f32, f32))```, which can make use of the rect function.

In the helper function:
- If flip_y is true, then use OpenGL uvs.
- If flip_y is false, then use DirectX, Metal, Vulkan uvs.

alexdevteam:
> It means, excluding Tileset and maybe some other types, that the user only needs to talk to the facade types rather than sometimes talking to Type1<'map> and sometimes talking to &'map Type2, which is a little inconsistent.

It would be inconsistent anyways. The lifetime of the suggested Tile facade would be of `'tileset`, not map. But yes, the interface is very inconsistent right now, specially in terms of types and properties.

bjorn:
> Add a function for obtaining the UV of a tile given its ID.

Just to mention, personally I would find it more suitable to pre-calculate and store the UVs (or rather, the pixel location) of each tile and store it in the `TileData`. That way the rendering loop doesn't need to keep calculating this based on the margin and spacing and such. But, I guess with `TileData` wrapped in the public API, it would be possible to add that later as well.

aleokdev:
Postponing to 0.11 since the change is no longer breaking.

PR #277, opened 2023-08-20
Add support for tile image rects in image collection tilesets

aleokdev:
Wasn't sure whether to put these members inside of TileData or put them in an external struct. Finally I ended up creating a new structure to hold them, but feel free to discuss against it.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 ### Added
+- Support for Tile `x`, `y`, `width` and `height` attributes related to image rects. (#277)
 - New `ImageLayer` `repeat_x`/`y` fields are now parsed from map image layers.
 
 ## [0.14.0]
--- a/src/tile.rs
+++ b/src/tile.rs
@@ -15,6 +15,20 @@ use crate::{
 /// A tile ID, local to a tileset.
 pub type TileId = u32;
 
+/// Indicates which part of an image should be used for a certain tile.
+/// Corresponds to the tile's `x`, `y`, `width` and `height` attributes.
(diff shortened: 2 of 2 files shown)


bjorn:
I know in the file format this information is currently only used for image collection tilesets. However, internally Tiled also uses it when tileset images are used. So I wonder if it would make sense to not make the `image_rect` optional, since it would actually be useful to provide this information in that context as well (see #166).

The long-term plan for storing each tile's image rect is to also be able to work with a texture atlas with varying tile sizes, and to improve the handling of expanding the image (when storing each tile's rect, image can be expanded at right and bottom sides without invalidating existing tiles).

Whether this should be a separate struct or not... I don't have a strong opinion. In Tiled I used a `QRect` just because I had it available and it makes it convenient to pass around. If Rust had a generic rectangle type it might make sense to use it instead, but the `ImageRect` struct is fine too.

aleokdev:
About calculating the image_rect at load time: To my understanding, the `<image>` tag inside the tileset may appear after any `<tile>` tag, so the image is not available when loading (Thus we cannot calculate the image rect). Maybe we could just add an image_rect fn to `Tile` apart from the member in its `TileData` which returns either the stored image_rect or calculates it from the parent tileset.

ElnuDev:
Any updates on this? This would be nice to have, as (to my knowledge) tile image rects are the way you're supposed to slice multiple sprites that are packed into a single spritesheet for multiple objects. Any easy workaround for now is just to copy all of the rect parameters as custom properties, but it is a bit clunky.

bjorn:
Right, I guess an optional `image_rect` works fine and we can indeed later add a function to `Tile` that will return this when it is set, or dynamically calculate it as suggested by @aleokdev. This can be merged as-is!

aleokdev:
Great, will try to merge today after work.

aleokdev:
Sorry for the delay once again.

bjorn (approved):

PR #278, opened 2023-08-31
Add content to text shape, fix kerning default value

aleokdev:
Somehow when adding text objects I forgot the most important thing of them: Their content. So there is currently no way to see the content of an object with the Text shape. I'm sorry for that. Sadly this change is breaking so it won't be featured until 0.12, but I'm considering releasing it just because of this.
Also fixes `kerning`'s default value, which should be `true`. (https://docs.mapeditor.org/en/latest/reference/tmx-map-format/#text)

### DIFF
--- a/src/objects.rs
+++ b/src/objects.rs
@@ -147,6 +147,8 @@ pub enum ObjectShape {
         kerning: bool,
         halign: HorizontalAlignment,
         valign: VerticalAlignment,
+        /// The actual text content of this object.
+        contents: String,
     },
 }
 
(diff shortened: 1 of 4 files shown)


src/objects.rs
```diff
@@ -449,6 +465,7 @@ impl ObjectData {
             kerning,
             halign,
             valign,
+            contents,
```
bjorn: I wonder if we could just call this `text`? I realize this may be stored without explicit name in the XML format, but `text` would match the property name used in the JSON format: https://doc.mapeditor.org/en/stable/reference/json-map-format/#text

aleokdev:
TODO: Add width+height attrs

aleokdev:
Done

bjorn (approved):

Issue #47, opened 2018-10-12
wasm support


aep:
flate2 doesnt suport wasm.

https://github.com/alexcrichton/flate2-rs/issues/161

should be easy to replace with  https://crates.io/crates/deflate

PR #281, opened 2023-12-27
replace libflate with flate2

quininer:
Initially `rs-tiled` used `flate2` but switched to `libflate` because `flate2` didn't support wasm https://github.com/mapeditor/rs-tiled/issues/47 , but `flate2` has been supporting wasm via `miniz_oxide` for a while now https://github.com/rust-lang/flate2-rs/issues/161.

This reduces dependency duplicated by other crates and has performance benefits.


bjorn (dismissed):
Looks fine to me!

aleokdev:
Could you please update `CHANGELOG.md` as well?

quininer:
I updated the changelog and version.

aleokdev (approved):
Thank you very much!

PR #282, opened 2024-02-18
Support for class properties

Deukhoofd:
Adds support for the new class properties introduced in Tiled 1.8, as described [here](https://doc.mapeditor.org/en/stable/reference/tmx-map-format/#property).

I had to write some slightly hacky code to work around the fact that its child tag is optional, which would break the parser when it tried to parse it anyway.

The new property value ClassValue differs from the other PropertyValues in that its a struct enum variant, instead of a tuple one. I did this due to the additional property type that is included in it. Having it as a struct variant feels cleaner to me than having it as a two element tuple.

### DIFF
--- a/assets/tiled_class_property.tmx
+++ b/assets/tiled_class_property.tmx
@@ -1,5 +1,5 @@
 <?xml version="1.0" encoding="UTF-8"?>
-<map version="1.5" tiledversion="1.7.2" orientation="orthogonal" renderorder="right-down" width="2" height="2" tilewidth="32" tileheight="32" infinite="0" nextlayerid="3" nextobjectid="4">
+<map version="1.10" tiledversion="1.10.2" orientation="orthogonal" renderorder="right-down" width="2" height="2" tilewidth="32" tileheight="32" infinite="0" nextlayerid="3" nextobjectid="4">
  <layer id="1" name="Tile Layer 1" width="2" height="2">
   <data encoding="csv">
 0,0,
@@ -11,7 +11,7 @@
    <properties>
     <property name="class property" type="class" propertytype="test_type">
      <properties>
-      <property name="test_property_1" value="3" type="int"/>
+      <property name="test_property_1" type="int" value="3"/>
      </properties>
     </property>
     <property name="empty property" type="class" propertytype="empty_type"/>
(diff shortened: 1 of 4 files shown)


aleokdev:
Thanks for the PR! I'll look at this on PC whenever I have time

assets/tiled_class_property.tmx
```diff
@@ -0,0 +1,21 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<map version="1.5" tiledversion="1.7.2" orientation="orthogonal" renderorder="right-down" width="2" height="2" tilewidth="32" tileheight="32" infinite="0" nextlayerid="3" nextobjectid="4">
```
aleokdev: Wasn't `class` added on Tiled 1.8? How was this file made?

aleokdev (dismissed):
Looks fine to me, I'll let @bjorn also take a look at it

assets/tiled_class_property.tmx
Deukhoofd: I modified the tiled_object_property.tmx using values from a map file of my own. Generating a full file for testing through Tiled would probably have been cleaner.

src/properties.rs
```diff
@@ -135,15 +143,31 @@ pub(crate) fn parse_properties(
...
+                // Class properties will have their member values stored in a nested <properties>
+                // element. Only the actually set members are saved. When no members have been set
+                // the properties element is left out entirely.
+                let properties = if has_properties_tag_next(parser) {
+                    parse_properties(parser)?
+                } else {
+                    HashMap::new()
+                };
+                p.insert(k, PropertyValue::ClassValue {
+                    property_type: p_t.unwrap_or_else(|| "class".to_owned()),
```
bjorn, line 166: I think it would be better to default to an empty string than to "class".

bjorn: I think it would be better to default to an empty string than to "class".
```suggestion
                    property_type: p_t.unwrap_or_default(),
```

assets/tiled_class_property.tmx
bjorn: I guess we can leave it as-is now, since it's not relevant for the test.

bjorn (commented):
Looks good to me too! I had just one suggestion.

bjorn (dismissed):

aleokdev:
Only thing missing are the changes in the CHANGELOG.md file, after that it's good to go! Thanks!

bjorn (dismissed):

aleokdev:
I forgot this is a breaking change (adds new variant to PropertyValue), so it should be based against `next`, not current, sorry!

bjorn:
@Deukhoofd I've tried to change the base branch of the PR, but it's a bit tricky, since you've created it directly from your `current` branch. Unfortunately, it is not possible to change the source branch of a PR, so either you'd need to temporary point your `current` to `next` or we need a new PR.

Or we could just use this PR to update the `next` branch with the other changes on `current`... though that's a bit messy.

aleokdev:
Couldn't these changes be rebased on next?

PR #283, opened 2024-02-26
Support for class properties

Deukhoofd:
Adds support for the new class properties introduced in Tiled 1.8, as described [here](https://doc.mapeditor.org/en/stable/reference/tmx-map-format/#property).

I had to write some slightly hacky code to work around the fact that its child tag is optional, which would break the parser when it tried to parse it anyway.

The new property value ClassValue differs from the other PropertyValues in that its a struct enum variant, instead of a tuple one. I did this due to the additional property type that is included in it. Having it as a struct variant feels cleaner to me than having it as a two element tuple.

Recreated PR from [here](https://github.com/mapeditor/rs-tiled/pull/282), to change target branch


Deukhoofd:
Recreated the PR [here](https://github.com/mapeditor/rs-tiled/pull/283)

aleokdev (approved):
Thanks a lot!

bjorn (approved):

PR #288, opened 2024-05-27
Fix u32 parse panic in decode_csv

smackysnacks:
This PR fixes an unhandled parsing error in `decode_csv` found with `cargo fuzz`. The following input no longer causes a panic:

```
<?xml version="1.0" encoding="UTF-8"?>
<map version="1.4" tiledversion="1.4.0" orientation="orthogonal" renderorder="right-down" width="100" height="100" tilewidth="32" tileheight="32" infinite="0" backgroundcolor="#ff00ff" nextlayerid="3" nextobjectid="5">
 <tileset firstgid="1" name="tilesheet" tilewidth="32" tileheight="32" tilecount="84" columns="14">
  <properties>
   <property name="tileset property" value="tsp"/>
  </properties>
  <image source="tilesheet.png" width="448" height="192"/>
  <tile id="1">
   <properties>
    <property name="a tile property" value="123"/>
   </properties>
  </tile>
 </tileset>
 <layer id="1" name="Tile Layer 1" width="100" height="100">
  <properties>
   <property name="prop1" value="12"/>
   <property name="prop2" value="some text"/>
   <property name="prop3">Line 1
Line 2
Line 3,
  etc
   </property>
  </properties>
  <data encoding="csv">
35,35,35,35,35,33,33,33,33,33,33,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,,0,
0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,
 </objectgroup>
</map>
```

### DIFF
--- a/src/error.rs
+++ b/src/error.rs
@@ -12,6 +12,8 @@ pub enum Error {
     DecompressingError(std::io::Error),
     /// An error occured when decoding a base64 encoded dataset.
     Base64DecodingError(base64::DecodeError),
+    /// An error occurred when decoding a csv encoded dataset.
+    CsvDecodingError(String),
     /// An error occured when parsing a XML file, such as a TMX or TSX file.
     XmlDecodingError(xml::reader::Error),
     /// The XML stream ended before the document was fully parsed.
@@ -68,6 +70,7 @@ impl fmt::Display for Error {
             Error::MalformedAttributes(s) => write!(fmt, "{}", s),
             Error::DecompressingError(e) => write!(fmt, "{}", e),
             Error::Base64DecodingError(e) => write!(fmt, "{}", e),
+            Error::CsvDecodingError(e) => write!(fmt, "{}", e),
             Error::XmlDecodingError(e) => write!(fmt, "{}", e),
             Error::PrematureEnd(e) => write!(fmt, "{}", e),
             Error::PathIsNotFile => {
--- a/src/error.rs
+++ b/src/error.rs
@@ -1,20 +1,39 @@
+use std::num::ParseIntError;
 use std::{fmt, path::PathBuf};
 
-/// Errors which occured when parsing the file
+/// Errors that can occur while decoding csv data.
+#[derive(Clone, Debug, PartialEq, Eq)]
+#[non_exhaustive]
+pub enum CsvDecodingError {
+    /// An error occurred when parsing tile data from a csv encoded dataset.
+    TileDataParseError(ParseIntError),
+}
+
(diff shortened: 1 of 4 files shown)


src/error.rs
```diff
@@ -12,6 +12,8 @@ pub enum Error {
     DecompressingError(std::io::Error),
     /// An error occured when decoding a base64 encoded dataset.
     Base64DecodingError(base64::DecodeError),
+    /// An error occurred when decoding a csv encoded dataset.
+    CsvDecodingError(String),
```
aleokdev: I'd prefer that this were an error type so the user could handle it (or format it some other way) if necessary

aleokdev (commented):
Many thanks!

src/error.rs
```diff
@@ -12,6 +12,8 @@ pub enum Error {
     DecompressingError(std::io::Error),
     /// An error occured when decoding a base64 encoded dataset.
```
bjorn: Maybe fix the previous misspellings of "occured" while you're at it. :-)

src/error.rs
smackysnacks: I think we have several options here. One option would be

```rs
CsvDecodingError(Box<dyn std::error::Error + Send + Sync + 'static>)
```

Alternatively, we could do something like

```rs
/// Errors that can occur while decoding csv data.
#[derive(Clone, Debug, PartialEq, Eq)]
pub enum CsvError {
    TileDataParseError(ParseIntError)
}

/// Errors which occurred when parsing the file
#[derive(Debug)]
#[non_exhaustive]
pub enum Error {
<SNIP>
    /// An error occurred when decoding a csv encoded dataset.
    CsvDecodingError(CsvError),
```

Do you have a preference?

aleokdev: Second one looks better IMO: Lets you look into the error, like in the other enum variants.

src/error.rs
smackysnacks: Done :)

aleokdev (approved):
LGTM!

bjorn (approved):

aleokdev:
I forgot to check for the changelog again x_x I'll update it when the version gets released I guess

bjorn:
@aleokdev Heh, we should probably look into adding a workflow warning about this. :D

aleokdev:
I have one set up for another project as a GH action, I'll look into importing it whenever possible if you'd like.

PR #297, opened 2024-06-12
Merge `next` with `current`

aleokdev:
Merge `current` into `next`. Let's see if I did it right this time...


bjorn (dismissed):

bjorn:
If there are no conflicts I think you could open a PR to do the merge directly, but if you had to resolve conflicts then I guess the right way is to merge locally and then open a PR. Note that you can now likely push the merge commit directly (which should automatically close this PR as merged) instead of doing it through GitHub, otherwise it would create another merge commit.

bjorn (approved):

PR #298, opened 2024-06-14
Update SFML

aleokdev:
Update SFML dev dependency for the SFML example

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -11,7 +11,7 @@ env:
 
 jobs:
   build:
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-24.04 # Version 24 required for last version of CSFML
 
     steps:
     - name: Checkout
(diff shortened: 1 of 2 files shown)


aleokdev:
Was expecting the CI fail. So the CI works with 0.20.0 because its CSFML package is outdated, however it fails on my system because I have the latest version installed.

bjorn:
> So the CI works with 0.20.0 because its CSFML package is outdated, however it fails on my system because I have the latest version installed.

I guess it's not so important whether we run CI on Ubuntu 22.04 or 24.04, so we could at the same time update it to run on `ubuntu-24.04` (currently in beta) instead of `ubuntu-latest`.

aleokdev:
Where did the `docs` workflow go? It's on the `next` branch, at least when viewed from the GitHub file browser, but I can't find it locally or in this PR

bjorn:
> Where did the `docs` workflow go? It's on the `next` branch, at least when viewed from the GitHub file browser, but I can't find it locally or in this PR

It's because this PR is based on `next` from before `current` was merged into it:

image

(it was added in the "Fix doc links" PR)

aleokdev:
Ah, right, thanks! Okay, I'll rebase just to make sure the `docs` workflow also gets the same treatment (although I don't think it needs it)

bjorn (approved):

PR #303, opened 2024-06-21
Add source members to `Tileset`, `Map` and `Template`

### LINKED ISSUE #263: Allow obtaining the path of tilemaps that have been loaded
I would like to get the path of the tileset used at the map.
But it does look like [Tileset](https://docs.rs/tiled/0.11.0/tiled/struct.Tileset.html ) does not provide this option.

aleokdev:
Closes #263. Adds a `source` member so users can obtain where stuff has been loaded from.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased (0.13.0)]
 ### Added
-- A way to obtain where tilesets, templates and maps have been loaded from
+- Added a `source` member to `Tileset`, `Map` and `Template`, which stores the resource path they have been loaded from.
 
 ## [0.12.0]
 ### Added
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,10 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [Unreleased (0.13.0)]
+### Added
+- A way to obtain where tilesets, templates and maps have been loaded from
+
 ## [0.12.0]
 ### Added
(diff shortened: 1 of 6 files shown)


CHANGELOG.md
```diff
@@ -5,6 +5,10 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [Unreleased (0.13.0)]
+### Added
+- A way to obtain where tilesets, templates and maps have been loaded from
```
bjorn: I think your commit title is more informative than your changelog entry. I'd suggest:
```suggestion
- Added a `source` member to `Tileset`, `Map` and `Template`, which stores the resource path they have been loaded from.
```

src/tileset.rs
```diff
@@ -17,6 +17,10 @@ pub use wangset::*;
 /// Also see the [TMX docs](https://doc.mapeditor.org/en/stable/reference/tmx-map-format/#tileset).
 #[derive(Debug, PartialEq, Clone)]
 pub struct Tileset {
+    /// The path first used in a [`ResourceReader`] to load this tileset.
+    ///
+    /// For embedded tilesets, this path will be the same as the template or map's source.
```
bjorn, line 22: That's an interesting choice. I might have made it an `Option<PathBuf>` and only set it for non-embedded tilesets. But maybe storing the original file path is actually more useful.

bjorn (commented):
It doesn't even look that hacky. :-)

CHANGELOG.md
aleokdev: I forgot to finish the changelog message 😅

src/tileset.rs
aleokdev: Yeah, thought about it, but storing the original source looks pretty useful. It's true that it's harder to know whether it's embedded or not but that's kinda an internal detail anyway

bjorn (dismissed):

aleokdev:
Sorry, wrong base

bjorn (approved):

PR #305, opened 2024-08-03
Implement `std::fmt::Debug` for `Map` manually

### LINKED ISSUE #138: Implement a proper Debug for Map / TileLayer
So we don't get bombarded with this when testing:
Screenshot_20220202_181259

erayerdin:
Closes #138 
Self-explanatory

### DIFF
--- a/src/map.rs
+++ b/src/map.rs
@@ -19,7 +19,7 @@ pub(crate) struct MapTilesetGid {
 }
 
 /// All Tiled map files will be parsed into this. Holds all the layers and tilesets.
-#[derive(PartialEq, Clone, Debug)]
+#[derive(PartialEq, Clone)]
 pub struct Map {
     version: String,
     /// The way tiles are laid out in the map.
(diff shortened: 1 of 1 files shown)


aleokdev (approved):
Sadly this means that we'll need to update `Debug` every time a member changes in `Map`, but it's better than what we had before IMO. Thanks!

aleokdev:
Didn't see this was a draft -- Are you planning to add anything else?

src/map.rs
```diff
@@ -65,6 +65,27 @@ pub struct Map {
...
+            .field("version", &self.version)
+            .field("orientation", &self.orientation)
+            .field("width", &self.width)
+            .field("height", &self.height)
+            .field("tile_width", &self.tile_width)
+            .field("tile_height", &self.tile_height)
+            .field("stagger_axis", &self.stagger_axis)
+            .field("stagger_index", &self.stagger_index)
+            .field("tilesets", &format!("{} tilesets", self.tilesets.len()))
+            .field("layers", &format!("{} layers", self.layers.len()))
```
erayerdin, line 80: Instead of debug-printing all tilesets and layers, we only debug-print the length of `self.tilesets` and `self.layers`.

I don't know if this is a viable approach as it calls `len` and allocates `String` with `format!`, which would be not as performant as desired. On the other hand, we're talking about `Debug` here, which should/must be only relied upon debug builds, not release builds (only logs maybe).

erayerdin:
> Didn't see this was a draft -- Are you planning to add anything else?

I was checking `TileLayer`, `FiniteTileLayer`, `InfiniteTileLayer` and `map_wrapper!` macro to see how I could implement `Debug` for these but it seems they are already done.

erayerdin:
@aleokdev [This](https://mcarton.github.io/rust-derivative/latest/Debug.html) seems like it would not require rewriting the whole implementation for `Map` ([this one](https://mcarton.github.io/rust-derivative/latest/Debug.html#format-with) specifically). This, however, would add extra to compile time as `derivative` heavily relies on macros.

If you find this viable, I could go down this path. :)

aleokdev:
It's ok, we'll do it manually - Same thing we have done for `Error` (we could've used `thiserror` but chose not to)

erayerdin:
Well, I'm going to check `TileLayer` stuff after I wake up. Thank you for your time. :)

erayerdin:
@aleokdev 

I've checked the `TileLayer`. Apparently, it is an enum with `FiniteTileLayer` and `InfiniteTileLayer`, both of which implement `Debug` manually (well, at least their inner data `FiniteTileLayerData` and `InfiniteTileLayerData` do), and they do not seem to include `Vec`s to their debug-print.

https://github.com/mapeditor/rs-tiled/blob/f86e079fc2ca9de362ed576d2ae8c21636bf479d/src/layers/tile/finite.rs#L19-L26

https://github.com/mapeditor/rs-tiled/blob/f86e079fc2ca9de362ed576d2ae8c21636bf479d/src/layers/tile/infinite.rs#L18-L22

So, I think it's good to go. Feel free to merge. :beer:

aleokdev:
Great, thank you!

PR #306, opened 2024-08-18
Map::layers() doesn't really list ALL layers - added a comment about it

HughHoyland:
Took me a while to realize when I tried to load one especially large map.

I guess layers API needs a bit of improvement, but let's start with a comment.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,10 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [Unreleased]
+### Changed
+- Improved documentation on `Map::layers` and `Map::get_layer`. (#306)
+
 ## [0.12.0]
 ### Added
 - Add `text`, `width` and `height` members to `ObjectShape::Text`. (#278)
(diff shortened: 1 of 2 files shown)


HughHoyland:
I see an even bigger problem.

While `Layer<'a>` is supposed to live as long as the `Map`, getting `Layer` out of a `LayerType::Group(GroupLayer)` only gives one a `Layer<'a>` where `'a` is the lifetime of the local copy of `GroupLayer`.

Which means it cannot be passed outside of a function.

It is probably safe to cast... in an unsafe block.

aleokdev:
> I see an even bigger problem.
> 
> While `Layer<'a>` is supposed to live as long as the `Map`, getting `Layer` out of a `LayerType::Group(GroupLayer)` only gives one a `Layer<'a>` where `'a` is the lifetime of the local copy of `GroupLayer`.

This is a bug. I see the problem: `get_layer` returns `Layer<'self>` instead of `Layer<'map> + 'map`. Has happened elsewhere before. Sorry about that, I'll submit a fix ASAP.

aleokdev (commented):
Looks good, thanks!
Please also add your changes to the `CHANGELOG.md` file under a new section:
```
## [Unreleased]
### Changed
- Improved documentation on `Map::layers` and `Map::get_layer`. (#306)
```

HughHoyland:
Added the changelog.

aleokdev (approved):
Thank you!

PR #309, opened 2024-08-19
Prepare for 0.12.1

aleokdev:
Prepare for a new crates.io release.

Version will be released shortly after merging this PR.


bjorn (approved):
I guess the lifetime fix can be important enough for people, and since no further fixes are planned, it makes sense not to wait with pushing this out.

PR #311, opened 2024-09-03
Fixed template instance size and position overrides

### LINKED ISSUE #310: Template property override
I have an ellipse object that triggers some functionality when the player enters the zone defined by the ellipse shape. I have some custom properties for this object so I have converted it into a template to reuse in many places.

The issue I'm having is that when I try to resize the ellipse, the new width and height is not in the `ObjectData` anywhere. It only gives me the width and height that is defined in the template file. Perhaps I am missing something but I couldn't find anything in the docs about this.

Here is the object from my .tmx file:
```xml
<object id="8" template="../templates/Trigger.tx" type="Trigger" x="617" y="5565" width="100" height="100">
```

And here is the template:
```xml
<template>
 <object type="Trigger" width="111" height="289">
<!-- custom properties -->
  <ellipse/>
 </object>
</template>
```

In this case the `ObjectData` only contains a width of 111 and height of 289 but I would expect it to have a width and height of 100. Is this the intended behaviour? I have another similar template with the same issue but it is a rectangle.

I am new to Tiled so let me know if there is a better way to go about this.

bjorn:
Template instances of rectangle, ellipse and text objects can override the width and height of the template object. This is now taken into account when reading out the attributes and when inheriting the shape from the template.

Template instances of point objects are now also correctly storing the instance position in the Point shape, instead of the position of the template object.

Closes #310

### DIFF
--- a/assets/tiled_object_template.tmx
+++ b/assets/tiled_object_template.tmx
@@ -1,5 +1,5 @@
 <?xml version="1.0" encoding="UTF-8"?>
-<map version="1.4" tiledversion="1.4.2" orientation="orthogonal" renderorder="right-down" width="3" height="3" tilewidth="32" tileheight="32" infinite="0" nextlayerid="3" nextobjectid="3">
+<map version="1.11" tiledversion="1.11.0" orientation="orthogonal" renderorder="right-down" width="3" height="3" tilewidth="32" tileheight="32" infinite="0" nextlayerid="3" nextobjectid="4">
  <tileset firstgid="1" source="tilesheet.tsx"/>
  <layer id="1" name="Tile Layer 1" width="3" height="3">
   <data encoding="csv">
@@ -14,5 +14,6 @@
    </properties>
   </object>
   <object id="2" gid="45" x="0" y="32" width="32" height="32"/>
+  <object id="3" template="tiled_object_template.tx" x="0" y="64" width="64" height="32"/>
  </objectgroup>
 </map>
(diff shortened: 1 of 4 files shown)


bjorn:
Whoops, I should have opened this PR from my personal fork.

bjorn:
I guess it would be good to add some kind of test for overriding size in a template instance...

aleokdev (dismissed):
Looks great!

bjorn:
@aleokdev Thanks! Most annoying was the handling of the Text objects, heh. What a bunch of boilerplate, which I don't think I could avoid thanks to it being an enum variant rather than a plain struct.

I decided to add a small test to verify the behavior.

assets/tiled_object_template.tmx
```diff
@@ -1,5 +1,5 @@
 <?xml version="1.0" encoding="UTF-8"?>
-<map version="1.4" tiledversion="1.4.2" orientation="orthogonal" renderorder="right-down" width="3" height="3" tilewidth="32" tileheight="32" infinite="0" nextlayerid="3" nextobjectid="3">
```
aleokdev, line 2: Wow, it's been a couple years heh

aleokdev (approved):

assets/tiled_object_template.tmx
bjorn: Yeah, though even at the time this file was added, Tiled 1.8 was already out. So the file appeared older than it actually was. :-)

PR #320, opened 2024-12-16
World File Parsing

JtotheThree:
Adds feature to parse Tiled World format.

Includes the ability to parse patterns as well (I think this will cause oddities with the `ResourceReader` though if the other files aren't preloaded into the reader). 

serde and serde_json have been added as dependencies.

I tried adding an option to pre-parse all of the TMX files referenced in the world file, but I ran into the same issue which I think the patterns might have as well, the files won't exist in the "folder" if they weren't already loaded into the ResourceReader.

I think loading all the TMX files might make sense to expect it to happen downstream.

The `ResourceReader` concept is a bit new to me, so I'm going to open this PR and hopefully someone will have advice on which direction to take it. I'm also not using the cache in the World parse_xml, I'm not sure if it's needed?


src/world.rs
```diff
@@ -0,0 +1,214 @@
...
+/// # use tiled::Loader;
+/// #
+/// # fn main() {
+/// #    let mut loader = Loader::new();
+/// #    let world = loader.load_world("assets/world/world_basic.world").unwrap();
+/// #    
+/// #    for map in world.maps.unwrap() {
+/// #        println!("Map: {:?}", map);
+/// #    }
+/// # }
```
aleokdev: `#` makes the line invisible in docs -- It's used for initializing stuff that the user doesn't care about or for defining dummy variables/functions so that the example can be compiled and run. Here however it'll render the entire block as empty in the docs, watch out!

aleokdev: Also, this doc comment belongs better to `Loader::load_world` instead

src/world.rs
```diff
@@ -0,0 +1,214 @@
...
+    let mut maps = Vec::new();
+
+    let parent_dir = path.parent().ok_or(Error::ResourceLoadingError {
+        path: path.to_owned(),
+        err: Box::new(std::io::Error::from(std::io::ErrorKind::NotFound)),
+    })?;
+
+    // There's no documentation on why "patterns" is a JSON array, so we'll just blast them into same maps list.
+    for pattern in patterns {
+        let files = fs::read_dir(parent_dir).map_err(|err| Error::ResourceLoadingError {
```
aleokdev: We shouldn't be using `std::fs` directly, since that won't work on web or places where the user has a virtual filesystem set up. We need to create a new method in ResourceReader instead which acts as an interface to `read_dir`.

Cargo.toml
```diff
@@ -36,6 +36,9 @@ base64 = "0.22.1"
 xml-rs = "0.8.4"
 zstd = { version = "0.13.1", optional = true, default-features = false }
 flate2 = "1.0.28"
+serde = "1.0.215"
+serde_json = "1.0.133"
+regex = "1.11.1"
```
aleokdev: These dependencies are pretty heavy, maybe we should feature-gate world file parsing (under a `world` feature perhaps)? It's not something that everyone uses, and it's in a completely different format to TMX files

src/world.rs
```diff
@@ -0,0 +1,214 @@
+use std::{
+    fs, io::Read, path::{Path, PathBuf}
+};
+
+use regex::Regex;
+use serde::Deserialize;
+
+use crate::{Error, Map, ResourceCache, ResourceReader};
+
+/// A World is a collection of maps and their layout in the game world.
```
aleokdev: I think this could be a bit more accurate: IMO it cannot be considered a collection of maps because the object does not actually contain any maps. It could be a collection of references to maps, though.

src/world.rs
```diff
@@ -0,0 +1,214 @@
...
+#[derive(Deserialize, PartialEq, Clone, Debug)]
+pub struct World {
+    /// The path first used in a ['ResourceReader'] to load this world.
```
aleokdev:
```suggestion
    /// The path first used in a [`ResourceReader`] to load this world.
```
Backticks are important to make the link to the relevant item in documentation.

src/world.rs
```diff
@@ -0,0 +1,214 @@
...
+    #[serde(skip_deserializing)]
+    pub source: PathBuf,
+    /// The maps present in this world.
+    pub maps: Option<Vec<WorldMap>>,
+    /// Optional regex pattern to load maps.
+    patterns: Option<Vec<WorldPattern>>,
```
aleokdev: I can understand why you have decided to make this member private, however as-is it doesn't really provide any meaningful information nor can it be used anywhere (I think). Maybe it's a good idea to just make them public, so that users can do stuff such as hot-reloading the pattern. Either that, or we get entirely rid of the patterns member. @bjorn What do you think?

src/world.rs
```diff
@@ -0,0 +1,214 @@
...
+
+                let x = match x
+                    .checked_mul(pattern.multiplier_x as i32)
+                    .and_then(|x| x.checked_add(pattern.offset_x))
+                {
+                    Some(x) => x,
+                    None => {
+                        return Err(Error::ResourceLoadingError {
+                            path: path.to_owned(),
+                            err: "Arithmetic Overflow on multiplierX and offsetX".into(),
```
aleokdev:
```suggestion
                            err: "Arithmetic overflow on multiplierX and offsetX".into(),
```

src/world.rs
```diff
@@ -0,0 +1,214 @@
...
+                    })?;
+                let y = match y
+                    .checked_mul(pattern.multiplier_y as i32)
+                    .and_then(|y| y.checked_add(pattern.offset_y))
+                {
+                    Some(y) => y,
+                    None => {
+                        return Err(Error::ResourceLoadingError {
+                            path: path.to_owned(),
+                            err: "Arithmetic Overflow on multiplierY and offsetY".into(),
```
aleokdev:
```suggestion
                            err: "Arithmetic overflow on multiplierY and offsetY".into(),
```

src/world.rs
```diff
@@ -0,0 +1,214 @@
...
+                };
+                Ok(WorldMap {
+                    filename,
+                    x,
+                    y,
+                    width: Some(pattern.multiplier_x),
+                    height: Some(pattern.multiplier_y),
+                })
+            })
+            .collect::<Vec<_>>();
```
aleokdev: I think you can collect into a `Result<Vec<_>, _>` here instead, and use the try operator right away. That way you can also exchange the for loop after this with just a `maps.append(files)`.

aleokdev (commented):
Thank you! I've got a few comments. Also please don't forget to update the `CHANGELOG.md`.

aleokdev:
Seems like some of the README examples broke, maybe because of some compiler update that changes how paths are handled with `include_str!`. Could you please fix them as well? Should be as easy as replacing `Ok(Cursor::new(include_bytes!("../assets/tiled_xml.tmx")))` by `Ok(Cursor::new(include_bytes!("assets/tiled_xml.tmx")))` and `Ok(std::io::Cursor::new(include_bytes!("../assets/tiled_csv.tmx")))` by `Ok(std::io::Cursor::new(include_bytes!("assets/tiled_csv.tmx")))`

src/world.rs
bjorn: Alternatively, we could expose a bit more low-level API, for example just to check whether a certain file name matches a pattern possibly returning an `Option<WorldMap>` in case there is a match, with the members filled in based on the capture (or `Option<(x: i32, y: i32)>` since that's the only new information).

Then the user can use `fs::read_dir` or whatever API they're using and check each file against the patterns.

src/world.rs
bjorn: I think it would make sense for this data to be public, especially if we just provide some convenience functions on `World` and `WorldPattern` to check whether a file name matches it.

src/world.rs
```diff
@@ -0,0 +1,214 @@
...
+/// If "patterns" key is present, it will attempt to build the maps list based on the regex patterns.
+fn parse_world_pattern(path: &Path, patterns: &Vec<WorldPattern>) -> Result<Vec<WorldMap>, Error> {
+    let mut maps = Vec::new();
+
+    let parent_dir = path.parent().ok_or(Error::ResourceLoadingError {
+        path: path.to_owned(),
+        err: Box::new(std::io::Error::from(std::io::ErrorKind::NotFound)),
+    })?;
+
+    // There's no documentation on why "patterns" is a JSON array, so we'll just blast them into same maps list.
```
bjorn: There are multiple patterns because one world can contain sets of maps at different offsets and/or with different naming patterns. It's not common, but I had to support at least one such a case. :-)

src/world.rs
JtotheThree: One question on the ResourceReader, if we add a read_dir implementation to the trait is that going to cause breakage with anyone providing their own implementation?

I think the alternative might be a better direction.

I'm leaning toward just providing the loaded pattern string back to anything downstream and let them deal with the regex or we provide helper function to test and parse a filename as Bjorn suggested.

Testing the folder for the user will create a hazard for any WASM user or even Bevy user if they don't have a virtual directory preloaded to give it and don't pay close attention to the documentation.

Cargo.toml
JtotheThree: I've added a crate feature "world" the will be in when I update the PR after addressing the other issue/topics. Thanks!

src/world.rs
```diff
@@ -0,0 +1,150 @@
...
+#[derive(Deserialize, PartialEq, Clone, Debug)]
+pub struct World {
+    /// The path first used in a [`ResourceReader`] to load this world.
+    #[serde(skip_deserializing)]
+    pub source: PathBuf,
+    /// The [`WorldMap`]s defined in the world file.
+    pub maps: Option<Vec<WorldMap>>,
+    /// Optional regex pattern to load maps.
+    pub patterns: Option<Vec<WorldPattern>>,
+    /// The type of world, which is arbitrary and set by the user.
```
bjorn: Hmm, actually there are not different types of worlds and this can't be set by the user. A world JSON file always has `"type": "world"` at the top-level just so it can be recognized as a world object.

So I think this member can be removed. The parser can either just ignore it or use it to verify that we're actually loading a world JSON object (I guess I'd go with the former since JSON seems to be commonly duck-typed, and Tiled ignores this value as well).

src/world.rs
```diff
@@ -0,0 +1,150 @@
...
+    #[serde(rename = "fileName")]
+    pub filename: String,
+    /// The x position of the map.
+    pub x: i32,
+    /// The y position of the map.
+    pub y: i32,
+    /// The optional width of the map.
+    pub width: Option<u32>,
+    /// The optional height of the map.
+    pub height: Option<u32>,
```
bjorn: In Tiled (and in JSON) these are just signed values, so I wonder if it's a good idea to apply an unsigned type here, even if we don't expect negative values.

src/world.rs
```diff
@@ -0,0 +1,150 @@
...
+    #[serde(rename = "multiplierX")]
+    pub multiplier_x: u32,
+    /// The multiplier for the y position.
+    #[serde(rename = "multiplierY")]
+    pub multiplier_y: u32,
+    /// The offset for the x position.
+    #[serde(rename = "offsetX")]
+    pub offset_x: i32,
+    /// The offset for the y position.
+    #[serde(rename = "offsetY")]
```
bjorn: Probably we could use `#[serde(rename_all = "camelCase")]` on the struct instead?

src/world.rs
```diff
@@ -0,0 +1,150 @@
...
+            x,
+            y,
+            width: None,
+            height: None,
+        })
+    }
+
+    /// Utility function to test a list of paths against the defined regexp field.
+    /// Returns a parsed list of WorldMaps from any matched filenames.
+    pub fn capture_paths(&self, paths: Vec<PathBuf>) -> Result<Vec<WorldMap>, Error> {
```
bjorn: Since this returns a single result rather than an array of results, does that mean that if any of the given paths do not turn out to be a valid entry for this pattern, you only get an error back?

In general I think people will want to iterate the files in a directory and check for each of them whether it is a valid entry in the world, so a helper function on `World` that checks a path against all patterns might be more useful.

Though as the one needing this feature you might have other ideas about this. :-)

src/world.rs
```diff
@@ -0,0 +1,150 @@
...
+    pub offset_x: i32,
+    /// The offset for the y position.
+    #[serde(rename = "offsetY")]
+    pub offset_y: i32,
+}
+
+impl WorldPattern {
+    /// Utility function to test a single path against the defined regexp field and returns a parsed WorldMap if it matches.
+    /// Returns none if the filename does not match the pattern.
+    pub fn capture_path(&self, path: &Path) -> Result<WorldMap, Error> {
```
bjorn: I'd suggest to call this `match_path`. In the end it's not the path that is captured, but the path is matched and the map's position is captured.

src/world.rs
```diff
@@ -0,0 +1,150 @@
...
+        let re = Regex::new(&self.regexp).unwrap();
```
bjorn: Since there's a cost involved in parsing the regex, it would be nice if `self.regexp` was already a `Regex` instead of a `String`. The `.unwrap()` should probably be changed to cause a world loading error.

I guess we could use https://crates.io/crates/serde_regex?

src/world.rs
```diff
@@ -0,0 +1,150 @@
...
+        })?;
+
+    let world: World = match serde_json::from_str(&world_string) {
+        Ok(world) => world,
+        Err(err) => {
+            return Err(Error::JsonDecodingError(err));
+        }
+    };
+
+    Ok(world)
```
bjorn: I guess this could be simplified to something like (untested):

```suggestion
    serde_json::from_str(&world_string)
        .map_err(|err| Error::JsonDecodingError(err)))
```

bjorn (changes requested):
Some further feedback from my side. Also I think this PR should target `next` instead of `current` (but that can be changed without opening a new PR).

src/world.rs
JtotheThree: Patterns has been made public.

src/world.rs
JtotheThree: Good suggestion. Change in latest commit.

src/world.rs
JtotheThree: I added the serde_regex crate to the `world` feature. I tried to avoid using it but the de-serialization rabbit hole would get to the point where most of the world.rs file would be dealing with that single variable.

src/world.rs
JtotheThree: I like match_. I did make another change and ended up with `match_filename` and switched the argument to &str. My thought process is that Tiled doesn't support paths in the regexp field according to the documentation. If we call it match_path it suggests to the user that the field supports paths. While it would technically work fine for rs-tiled if they loaded that file into Tile it would give them issues.

src/world.rs
JtotheThree: I've moved the pattern utilities to the `World` implementation and it now tests all defined patterns and parses based on the first matched pattern listed. From looking at the code in Tiled that's how it handles it as well.

src/world.rs
JtotheThree: Great suggestion. Changed in the latest commit.

src/world.rs
JtotheThree: Changed to i32 in the latest commit.

src/world.rs
JtotheThree: Latest commit handles multiple patterns and resolves the first matched pattern in the array that matches. I believe this is how Tiled handles it as well.

src/world.rs
JtotheThree: I've removed the world_type entirely.

JtotheThree:
Thanks for taking the time to provide thoughtful review. I haven't contributed much to open source yet but I would like to change that so I really appreciate it.

I think the last couple of things I'm not certain on.

`match_filename` vs `match_path`. I've provided reasoning in the comment above, but my reasoning is that Tiled doesn't support actual paths in the world files, the tmx files are expected to be relative to the world file.

I went with Result instead of Option since regular expressions can be really frustrating to work with. That way the user doesn't provide a list of files and just get None back. This has fallen a bit short with testing a filename against every pattern in the world file since it's to be expected that some of the patterns won't fit so it doesn't make sense to return why. I still think it's OK as we can return an error if the arithmetic overflows instead of panicking. The world file patterns are hand edited by the user so they are more prone to errors.

src/world.rs
bjorn: Hmm, actually the full paths are matched in Tiled, it's just that only files from the same directory as the world file will be tested on the regex. This means the position can be captured also from somewhere in the path rather than the file name, though I'm not sure that's useful to anybody.

So it doesn't matter much today, but I could imagine a world file could in the future be changed to be able to refer to maps in other directories, so I think `match_path` is in the end the better name.

JtotheThree: Converted back to match_path and match_paths. match_path takes AsRef<Path> so it's a better interface to work overall.

JtotheThree:
I'll look into rebasing to next.

bjorn:
> I'll look into rebasing to next.

I've changed the target branch to `next`. It looks like there are no further changes necessary. We were lucky this time because no new commits had been made to `current` yet that hadn't been already merged to `next`. :-)

About the latest version:

* I kind of liked that it was possible to match with a single `WorldPattern` against a path before, and had expected the `World::match_path` to just loop over the patterns and use the existing function. Even if you think such API isn't useful, I would still suggest doing this with a private function for cleaner code.

* Since [`Path::to_str`](https://doc.rust-lang.org/std/path/struct.Path.html#method.to_str) might involve a check on valid unicode, it would be good to do it only once per path when iterating the patterns.

I hope you're not growing tired of the back and forth with the reviews! I have a feeling these will be my last concerns, but let's see what @aleokdev will say. :-)

.github/workflows/rust.yml
```diff
@@ -27,10 +27,10 @@ jobs:
       run: sudo apt-get install -y libsfml-dev libcsfml-dev libasound2-dev libudev-dev
 
     - name: Build library
-      run: cargo build --lib --verbose
+      run: cargo build --lib --all-features --verbose
```
aleokdev, line 30: We should probably also have a build check without all features. However this could be improved later on using `cargo hack` to check with the powerset of features instead. So all good here for now.

Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "tiled"
-version = "0.13.0"
```
aleokdev, line 3: Need to update version on README as well

JtotheThree:
>     * I kind of liked that it was possible to match with a single `WorldPattern` against a path before, and had expected the `World::match_path` to just loop over the patterns and use the existing function. Even if you think such API isn't useful, I would still suggest doing this with a private function for cleaner code.

I think that's a good suggestion. I've added it back to the `WorldPattern` impl as public and `World` now utilizes it. 

>     * Since [`Path::to_str`](https://doc.rust-lang.org/std/path/struct.Path.html#method.to_str) might involve a check on valid unicode, it would be good to do it only once per path when iterating the patterns.

I added a single conversion before the iterative call.

`    pub fn match_path(&self, path: impl AsRef<Path>) -> Result<WorldMap, Error> {
        let path_str = path.as_ref().to_str().expect("obtaining valid UTF-8 path");

        if let Some(patterns) = &self.patterns {
            for pattern in patterns {
                match pattern.match_path(path_str) {`

`WorldPattern::match_path` makes the same conversion though. My understanding is that since a &str is passed in the `World::match_path` call that it'll fall through without the valid UTF-8 check so it's still an improvement.


> I hope you're not growing tired of the back and forth with the reviews! I have a feeling these will be my last concerns, but let's see what @aleokdev will say. :-)

I don't mind at all. I don't have a lot of experience writing generalized public libraries so this is very educational for me and I appreciate you making the effort for review.

src/loader.rs
```diff
@@ -182,6 +185,18 @@ impl<Cache: ResourceCache, Reader: ResourceReader> Loader<Cache, Reader> {
         crate::parse::xml::parse_tileset(path.as_ref(), &mut self.reader, &mut self.cache)
     }
 
+    #[cfg(feature = "world")]
+    /// Parses a file hopefully containing a Tiled world.
+    ///
+    /// The returned [`World`] provides the deserialized data from the world file. It does not load
+    /// any maps or tilesets.
+    /// ## Note
+    /// The ['WorldPattern`] struct provides [`WorldPattern::capture_path`] and [`WorldPattern::capture_paths`]
```
bjorn: Outdated docs.

src/world.rs
```diff
@@ -0,0 +1,186 @@
...
+
+/// A World is a list of maps files or regex patterns that define a layout of TMX maps.
+/// You can use the loader to further load the maps defined by the world.
+#[derive(Deserialize, PartialEq, Clone, Debug)]
+pub struct World {
+    /// The path first used in a [`ResourceReader`] to load this world.
+    #[serde(skip_deserializing)]
+    pub source: PathBuf,
+    /// The [`WorldMap`]s defined by the world file.
+    pub maps: Option<Vec<WorldMap>>,
```
bjorn: Just wondering, but is it useful for this to be an `Option`? Alternatively I think we could just add:
```rust
#[serde(skip_serializing_if = "Vec::is_empty", default)]
```
Since there is basically no difference between an absent array or an empty one, and this could simplify access a little. Same for `patterns` below.

src/world.rs
```diff
@@ -0,0 +1,186 @@
...
+            && self.offset_y == other.offset_y
+            && self.regexp.to_string() == other.regexp.to_string()
+    }
+}
+
+impl WorldPattern {
+    /// Utility function to test a path against this pattern.
+    /// Returns a parsed [`WorldMap`] on the first matched pattern or an error if no patterns match.
+    pub fn match_path(&self, path: impl AsRef<Path>) -> Result<WorldMap, Error> {
+        let path_str = path.as_ref().to_str().expect("obtaining valid UTF-8 path");
```
bjorn, line 104: I somewhat doubt the compiler will be able to optimize this check away after it already happened in `World::match_path`. However, since both functions are public API I think they should take a consistent type. Maybe @aleokdev can help decide between `&str` and `AsRef<Path>`?

bjorn (commented):
Looking great now! Just left a few small comments.

src/world.rs
aleokdev: What I've seen in other crates and read in the amazing Rust for Rustaceans book by Jon Gjenset is that, with functions that take a generic parameter which is later monomorphized, two functions are created, one which is public API and exposes the generic parameter (in our case `impl AsRef<Path>` and another which is private and defined either inside the public function's body or alongside it which just takes the type it is converted to (in our case `&str`). That way the function body only gets defined once on the binary, and the only thing that is copied per type input in the generic function is the conversion to &str.

So I think it could be great to apply here. Have the public API function which just calls a `pub(crate) fn match_path_impl` which takes in &str, and use that from World too.

Cargo.toml
JtotheThree: Fixed the readme

JtotheThree:
Added the match_path_impl pub(crate) function and updated the public interfaces to use that internally.
Update the readme to reflect a version change.

src/world.rs
JtotheThree: Changed to empty vecs. It's a lot more ergonomic using it from another crate now as it reduces the need to unwrap the option.

JtotheThree:
Ok, I think that's everything.

There's some oddities with the readme include_bytes paths. On Mac and on the automated build the paths work as "assets/" but on my Linux machine it wants them back as "../assets". I run the tests from CLI so nothing like vscode would interfere with that.

src/world.rs
```diff
@@ -0,0 +1,190 @@
...
+
+use crate::{Error, ResourceReader};
+
+/// A World is a list of maps files or regex patterns that define a layout of TMX maps.
+/// You can use the loader to further load the maps defined by the world.
+#[derive(Deserialize, PartialEq, Clone, Debug)]
+pub struct World {
+    /// The path first used in a [`ResourceReader`] to load this world.
+    #[serde(skip_deserializing)]
+    pub source: PathBuf,
```
adrien-bon, line 17: Hi, sorry to interfere but it seems like this field is never set. At least in my testings it's always empty.
Would be nice to have the asset source as it is done for maps.

JtotheThree: Good catch. I've fixed that in the latest commit and made sure it's checked in the test.

Thanks!

bjorn (approved):
No further comments from my side. Thanks for this nice addition!

aleokdev (approved):

aleokdev:
Thank you!

adrien-bon:
Hi!

@bjorn, @aleokdev, 
would it be possible to release an official version of the crate with this change?

[bevy_ecs_tiled](https://github.com/adrien-bon/bevy_ecs_tiled) now depends on this `world` feature (thanks @JtotheThree for the contrib!) but I cannot release a version for my crate without a `rs-tiled` release.

Thanks!

aleokdev:
Sure thing, and sorry for not doing it earlier! Should be uploaded now

PR #322, opened 2025-02-22
Update `current`

aleokdev:
Remember to do merge from Git

PR #324, opened 2025-08-18
Parse repeatx/repeaty from Image layers

guysv:
This allows map renderers to respect the image layer repeatx/repeaty properties.

relates to https://github.com/adrien-bon/bevy_ecs_tiled/issues/123

### DIFF
--- a/src/layers/image.rs
+++ b/src/layers/image.rs
@@ -1,8 +1,10 @@
 use std::{collections::HashMap, path::Path};
 
+use xml::attribute::OwnedAttribute;
+
 use crate::{
     parse_properties,
-    util::{map_wrapper, parse_tag, XmlEventResult},
+    util::{get_attrs, map_wrapper, parse_tag, XmlEventResult},
     Error, Image, Properties, Result,
 };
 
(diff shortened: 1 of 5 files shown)


aleokdev:
Thank you for your contribution! Please run rustfmt on it and I'll try to check it tomorrow morning

src/layers/image.rs
```diff
@@ -11,18 +13,32 @@ use crate::{
 pub struct ImageLayerData {
     /// The single image this layer contains, if it exists.
     pub image: Option<Image>,
+    /// The layer's x repeat factor (0 = no repeat).
+    pub repeat_x: i32,
+    /// The layer's y repeat factor (0 = no repeat).
+    pub repeat_y: i32,
```
aleokdev: This should probably be optional as it was not introduced until Tiled 1.8, and the TMX docs don't mention that it has a default value

aleokdev (commented):
This seems to be breaking, so please base your changes against `next` instead. Thanks!

src/layers/image.rs
bjorn: Right, Tiled only writes out this attribute when it is set, a 0 is never written. Also, even though Tiled uses a "1" to indicate repetition in the TMX format, the appropriate type here should be bool (it's a bool internally and the JSON format just uses bool as well).

I'll add mentioning of the default to the docs.

bjorn: Ah, to clarify, I thought this was about the XML attributes. These properties should definitely not be optional here, but simply default to false.

bjorn (commented):
Looks great to me. Now it's only missing an update to `CHANGELOG.md`. :-)

guysv:
not sure bout the CI but my test pass 😬

aleokdev:
The behavior of the include macro might've changed at some point, but it's okay, I'll see if I can fix it in another branch

aleokdev:
Oh yeah nevermind, it's this: https://github.com/mapeditor/rs-tiled/pull/322#issuecomment-2678267025
Still have no clue what's triggering it

aleokdev:
> not sure bout the CI but my test pass 😬 

Really sorry about this. Could you rebase now? It should be fine after that.

aleokdev (approved):
LGTM, rebased and modified the changelog to fit our convention :)

aleokdev:
@guysv There you go!

guysv:
thanks. sorry i didn't get to it! busy weekend

aleokdev:
Yes of course, don't worry about it, I just did it because I was already at it with the CI fixes and whatnot :)

PR #326, opened 2025-08-25
Fix CI

aleokdev:
Tries to fix CI. Sorry for the delay.


aleokdev:
So I'm not exactly sure what is wrong with the CI. I know it's related to this: https://doc.rust-lang.org/edition-guide/rust-2024/rustdoc-nested-includes.html, which specifies that until edition 2024, what we were doing didn't even compile. The thing is, until a few months ago, it did. So I have no idea what triggers the issue. I've also tried rolling back the toolchain version but even that fails to fix the issue.

The only ways I see to fix this are:
- Remove the tests that use `include!` macros entirely from the README.
- Ignore them.
- Update the entire crate to edition 2024, which may affect some users and will take some effort to achieve.

I've chosen to ignore them, at least for now. This solution is not perfect but I think it will do for now, at least this way we'll be able to merge a few PRs.

bjorn (approved):
I think ignoring until switching to the 2024 edition is fine. 👍

PR #332, opened 2026-01-20
updated edition to 2024, fixed clippy lints

sanbox-irl:
This PR updates some code to pass clippy's standard lints and also updates the project to use Edition 2024. This happen in one PR because certain lint are only really fixable with features from the more recent compiler versions. We can actually use 1.92, for example, with edition 2018, but I think it makes sense to just be on the current edition as well.

Functionally, this PR increases the MSRV (which is unstated currently) to 1.62 at least,
which came out in mid-2022. This was required for deriving default on enumerations.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -42,7 +42,7 @@ base64 = "0.22.1"
 xml-rs = "0.8.4"
 zstd = { version = "0.13.1", optional = true, default-features = false }
 flate2 = "1.0.28"
-serde = { version = "1.0.216", optional = true }
+serde = { version = "1.0.216", features = ["derive"], optional = true }
 serde_json = { version = "1.0.133", optional = true }
 serde_regex = { version = "1.1.0", optional = true }
 regex = { version = "1.11.1", optional = true }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -8,8 +8,13 @@ repository = "https://github.com/mapeditor/rs-tiled"
 readme = "README.md"
 license = "MIT"
 authors = ["Matthew Hall <matthew@quickbeam.me.uk>"]
-edition = "2018"
+edition = "2024"
 include = ["src/**/*.rs", "README.md", "LICENSE", "CHANGELOG.md"]
+
+[lints]
+rust.rust_2018_idioms = "deny"
+clippy.too_many_arguments = "allow"
+
 
 [features]
 default = ["zstd"]
--- a/README.md
+++ b/README.md
@@ -22,16 +22,13 @@ The minimum supported TMX version is 0.13.
 ```rust
 use tiled::Loader;
 
-fn main() {
-    let mut loader = Loader::new();
-    let map = loader.load_tmx_map("assets/tiled_base64_external.tmx").unwrap();
-    println!("{:?}", map);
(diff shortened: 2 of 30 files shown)


Cargo.toml
```diff
@@ -37,7 +42,7 @@ base64 = "0.22.1"
 xml-rs = "0.8.4"
 zstd = { version = "0.13.1", optional = true, default-features = false }
 flate2 = "1.0.28"
-serde = { version = "1.0.216", optional = true }
+serde = { version = "1.0.216", features = ["derive"], optional = true }
```
aleokdev, line 45: Oops, I wonder how this worked before, as we are using derive macros...

README.md
```diff
@@ -71,9 +68,8 @@ The crate supports WASM, but since it does not currently support asynchronous lo
 tiled = { version = ".....", features = ["wasm"] }
 ```
 
-- Second, since you cannot use the filesystem as normally on the web, you cannot use `FilesystemResourceReader`. As such,
-you'll need to implement your own `ResourceReader`. This is a pretty simple task, as you just need to return anything
-that is `Read`able when given a path, e.g.:
+- Second, since you cannot use the filesystem as normally on the web, you cannot use `FilesystemResourceReader`. As such, you'll need to implement your own `ResourceReader`. This is a pretty simple task, as you just need to return anything that is `Read`able when given a path, e.g.:
+
```
aleokdev, line 72: I don't think this is a very relevant change

aleokdev (commented):
Looks good, just one nitpick...

Cargo.toml
sanbox-irl: In 2018 (and 2015...and maybe 2021, but god the years are blurring!), the resolver did more feature unification. By switching to 2024, we're now using the current resolver, which does less unification.

Previously, we were pulling in serde and also semver was pulling serde+derive, which unified to us just pulling in serde+derive, which is why it was working before.

README.md
sanbox-irl: I agree, but weirdly clippy does warn about this as well. I suspect it's because the README is included as a mod comment for `lib.rs`.

I don't actually feel very strongly about this, but it does technically render more reliably correctly without the extra breaks.

aleokdev:
@sanbox-irl CI is failing, could you look at that?

bjorn:
This should probably go to `next`?

sanbox-irl:
As per the discord conversation, perhaps `next` should be removed? I do not think the possibility of rare backported bug-fixes is worth the development complexity

bjorn:
> As per the discord conversation, perhaps `next` should be removed? I do not think the possibility of rare backported bug-fixes is worth the development complexity

Probably it should, but until we've made that step we choose `next` or `current` based on whether it breaks compatibility. :-)

aleokdev (approved):

aleokdev:
Merge conflicts? Next needs an update...

bjorn:
@sanbox-irl You did not give us permission to push to your branch. Can you rebase these changes onto `next` so that this can be merged?

Cargo.toml
bjorn: Is this still relevant after #339, or was this limited to the `world` feature?

bjorn:
Superseded by #347.

Issue #137, opened 2022-02-02
Consider switching from xml-rs to quick-xml


bjorn:
According to its own documentation, [`quick-xml`](https://crates.io/crates/quick-xml) can be 50 times faster than [`xml-rs`](https://crates.io/crates/xml-rs). Of course, `xml-rs` might be fast enough for most people, but I think it's worth looking into the required porting effort and doing some measurements.

bjorn:
This was implemented on `next` for the upcoming rs-tiled 0.16.

PR #333, opened 2026-01-20
Use quick-xml

bjorn:
See #137


bjorn:
Ok, I "merged" this by accident due to a wrong push. The push has been force-pushed corrected, but this PR can't be reopened so I'll open a new one. :-/

PR #334, opened 2026-01-21
Switch from xml-rs to quick-xml

### LINKED ISSUE #137: Consider switching from xml-rs to quick-xml

bjorn:
Reopening #333

This change switches from `xml-rs` to `quick-xml`, taking care to avoid needless allocations and copying where possible.

Closes #137

### DIFF
--- a/src/error.rs
+++ b/src/error.rs
@@ -56,7 +56,7 @@ pub enum Error {
     /// An error occurred when decoding a csv encoded dataset.
     CsvDecodingError(CsvDecodingError),
     /// An error occurred when parsing an XML file, such as a TMX or TSX file.
-    XmlDecodingError(xml::reader::Error),
+    XmlDecodingError(quick_xml::Error),
     #[cfg(feature = "world")]
     /// An error occurred when attempting to deserialize a JSON file.
     JsonDecodingError(serde_json::Error),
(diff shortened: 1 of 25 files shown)


src/error.rs
```diff
@@ -56,7 +56,7 @@ pub enum Error {
     /// An error occurred when decoding a csv encoded dataset.
     CsvDecodingError(CsvDecodingError),
     /// An error occurred when parsing an XML file, such as a TMX or TSX file.
-    XmlDecodingError(xml::reader::Error),
+    XmlDecodingError(quick_xml::Error),
```
aleokdev, line 59: Thing I probably forgot to do with the previous iteration: If we expose a third party type on this enum, we should also export it using `pub use quick_xml` so that the user can refer to it.

aleokdev (commented):
Needs updating the CHANGELOG :)

src/error.rs
bjorn: Where should that line go?

aleokdev (approved):

bjorn:
Thanks @aleokdev, but I might have preferred to squash that myself and reword the commit message to something shorter. Oh well, I guess it's fine like this too. :-)

And what about that unresolved thing about `pub use quick_xml`?

aleokdev:
> And what about that unresolved thing about `pub use quick_xml`?

I completely forgot about that, sorry x_x

Ideally we should put that on the crate root, so that it is available at `tiled::quick_xml`

bjorn:
> Ideally we should put that on the crate root, so that it is available at `tiled::quick_xml`

Alright, I'll let you create a PR for this. :-)

PR #337, opened 2026-02-22
Make SFML and GGEZ examples optional features so tests can run

garvittsingla:
# Make SFML and GGEZ examples optional 

Earlier : 
When we use to run tests sfml and ggez compiled as they were dev dependencies and have to be resolved for sure
As a result, running:
`cargo test` attempts to compile native graphics libraries (CSFML).
This fails in headless environments and requires contributors to install graphics libraries even when they only want to use the tiled library.

After :
Making them as a dependency  as  optional and marking them as a feature  the ggez and sfml will only be compiled when features are enabled

Making 2 features each one for them
```rust
[features]
default = ["zstd"]
wasm = ["zstd/wasm"]
world = ["regex", "serde", "serde_json", "serde_regex"]
sfml-example = ["dep:sfml"]
ggez-example = ["dep:ggez"]

```
Making them compile with required features forcing to compile when required

```rust
[[example]]
name = "sfml"
path = "examples/sfml/main.rs"
required-features = ["sfml-example"]

[[example]]
name = "ggez"
path = "examples/ggez/main.rs"
required-features = ["ggez-example"]
```

## How to run SFML and GGEZ examples
```bash
cargo run --example --ggez --features=ggez-example
cargo run --example --sfml --features=sfml-example
```

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -15,6 +15,8 @@ include = ["src/**/*.rs", "README.md", "LICENSE", "CHANGELOG.md"]
 default = ["zstd"]
 wasm = ["zstd/wasm"]
 world = ["regex", "serde", "serde_json", "serde_regex"]
+sfml-example = ["dep:sfml"]
+ggez-example = ["dep:ggez"]
 
 [lib]
 name = "tiled"
@@ -27,10 +29,12 @@ path = "examples/main.rs"
 [[example]]
 name = "sfml"
 path = "examples/sfml/main.rs"
+required-features = ["sfml-example"]
 
 [[example]]
 name = "ggez"
 path = "examples/ggez/main.rs"
+required-features = ["ggez-example"]
 
 [dependencies]
 base64 = "0.22.1"
@@ -41,10 +45,5 @@ serde = { version = "1.0.216", optional = true }
 serde_json = { version = "1.0.133", optional = true }
 serde_regex = { version = "1.1.0", optional = true }
 regex = { version = "1.11.1", optional = true }
-
-[dev-dependencies.sfml]
-version = "0.21.0"
-features = ["graphics"]
-
-[dev-dependencies.ggez]
-version = "0.9.3"
+sfml = { version = "0.21.0", features = ["graphics"], optional = true }
+ggez = { version = "0.9.3", optional = true }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -15,8 +15,8 @@ include = ["src/**/*.rs", "README.md", "LICENSE", "CHANGELOG.md"]
 default = ["zstd"]
 wasm = ["zstd/wasm"]
 world = ["regex", "serde", "serde_json", "serde_regex"]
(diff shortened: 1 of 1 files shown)


bjorn:
> How to run SFML and GGEZ examples

Hmm, it's really unfortunately that one would need to specify those dependencies explicitly just to run the examples.

@aleokdev Could you review this? I wonder if Cargo has any better way to do this.

garvittsingla:
@bjorn is there something that i need to change?

bjorn:
@garvittsingla Well, if this is the best we can do with current Cargo features, I think there should at least be a section added to the `README.md` about running the examples because it may not be obvious to some people how they can enable those needed features. I also wonder if the examples are still part of the CI? But I asked @aleokdev for their review because they're more familiar with Rust.

garvittsingla:
@bjorn okayyy, Let me know if i need to make a docs for it as well, so that it is clearly specified how to use the examples!

bjorn:
@garvittsingla I had a look through the cargo features and it seems more ergonomic to use `test = false` for the example targets to exclude them from `cargo test`. See the [target selection](https://doc.rust-lang.org/cargo/commands/cargo-test.html#target-selection) section. And we maybe should do the same thing for `cargo bench`, even though we have no benchmarks at all yet (would be nice to add).

So could you try whether this also works and adjust the PR accordingly?

```toml
[[example]]
name = "sfml"
path = "examples/sfml/main.rs"
test = false
bench = false

[[example]]
name = "ggez"
path = "examples/ggez/main.rs"
test = false
bench = false
```

That should make `cargo test` not build these nor their dependencies, while still allowing `cargo run --example sfml` to just work.

garvittsingla:
@bjorn i gave it a try to it but it is still compiling
In target selection it is written that it will not test by ```test=false``` but it will **compile** normally
and i tested it locally as well
i set the ```autoexamples = false``` said in some stackoverflow forums as well, still same output of still compiling

image


Please tell me furthermore , how to continue with this PR

bjorn:
@garvittsingla Indeed seems that because these dependencies end up in `dev-dependencies` they will always be built, even when limiting the scope of `cargo test` using for example `cargo test --tests`. That means your approach of using features is probably the only solution we have.

> Let me know if i need to make a docs for it as well, so that it is clearly specified how to use the examples!

I checked the behavior of cargo and it says:

```
bjorn@thor:rs-tiled$ cargo run --example sfml
error: target `sfml` in package `tiled` requires the features: `sfml-example`
Consider enabling them by passing, e.g., `--features="sfml-example"`
```

I think this is clear enough for people trying to run the tests, and there are no existing instructions for running the examples that need adjustment, so I guess we don't need to change anything else at this point. Thanks!

bjorn (approved):

bjorn:
> I also wonder if the examples are still part of the CI?

I checked this and they are, since CI runs the tests with `--all-features`.

PR #339, opened 2026-04-01
fix compile error when the `world` feature is enabled, but serde's `derive` feature isn't

tesselode:


bjorn:
Looks good, thanks! Could you also add an entry to the `CHANGELOG.md`?

bjorn (approved):

PR #340, opened 2026-07-03
add support for list properties (fixed merge conflicts)

AwfulToTheEar:


bjorn:
Unfortunately, this change targets the wrong branch, it needs to target `next` which is the branch for the upcoming 0.16 release. And since this branch was ported from `xml-rs` to `quick-xml` there will be some changes needed.

Also, the fix for nested lists looks a bit complicated, we shouldn't need to keep track of the depth and duplicate code in `parse_tag_with_depth` if we rather parse lists recursively. I'll have a go at this.

bjorn:
Alright, I've opened #341 with the combination your work ported onto `next`, closing this PR in favor of that one. Thanks for noticing the nesting issue, @AwfulToTheEar!

